PR #52, opened 2024-02-10
Chat interface

ege0zcan:
## Problem
This PR introduces a Chat interface that can be used for chat events. This PR defines 3 main interfaces which are:

```
onChatPrompt
onEndChat
onQuickAction
```
These interfaces allow server creators to provide request handlers for the basic chat actions. 

This PR also introduces additional interfaces that are currently marked as optional. These interfaces allow server creators to register notification handlers for chat events like `onTabChange`

## License


src/features/chat/constants.ts
```diff
@@ -0,0 +1,29 @@
+export interface ChatItemAction {
```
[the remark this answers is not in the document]
ege0zcan: They can be kept in the same file as well. 

constants are closer to the UI logic and are based on mynah-ui properties. types are where the LSP methods and extensions are defined. To me that separation is easier to see this way

src/runtimes/standalone.ts
```diff
@@ -193,7 +204,7 @@ export const standalone = (props: RuntimeProps) => {
         const credentialsProvider: CredentialsProvider = auth.getCredentialsProvider()
 
         // Initialize every Server
-        const disposables = props.servers.map(s => s({ credentialsProvider, lsp, workspace, telemetry, logging }))
+        const disposables = props.servers.map(s => s({ chat, credentialsProvider, lsp, workspace, telemetry, logging }))
```
volodkevych, line 207: This is a breaking change for servers, right?

ege0zcan: yes

src/features/chat/constants.ts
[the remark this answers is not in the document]
volodkevych: `constants` name is confusing IMO

src/features/chat/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    onEndChat: (handler: RequestHandler<EndChatParams, EndChatResult, void>) => void
+    onQuickAction: (handler: RequestHandler<QuickActionParams, ChatResult, void>) => void
+
+    // Notifications
+    onSendFeedback?: (handler: NotificationHandler<VoteParams>) => void
+    onReady?: (handler: NotificationHandler<void>) => void
+    onTabAdd?: (handler: NotificationHandler<TabEventParams>) => void
+    onTabChange?: (handler: NotificationHandler<TabEventParams>) => void
+    onTabRemove?: (handler: NotificationHandler<TabEventParams>) => void
+    onVote?: (handler: NotificationHandler<VoteParams>) => void
```
volodkevych, line 26: How would `onVote` be different from `onSendFeedback` given they use the same parameters type?

src/features/chat/types.ts
```diff
@@ -0,0 +1,54 @@
...
+export type EndChatParams = { tabId: string }
+export type EndChatResult = boolean
+export const endChatRequestType = new ProtocolRequestType<EndChatParams, EndChatResult, never, void, void>(
+    'aws/endChat'
+)
+
+export interface QuickActionParams {
+    tabId: string
+    quickAction: string
+    prompt?: string
```
volodkevych, line 40: nit: would it make sense to reuse `ChatPrompt` type here?

src/features/chat/index.ts
ege0zcan: They share the same input shape but correspond to different events in the UI. I will create an explicit interface for feedback params that will be same as voteparams for the time being. This should allow future flexibility and explicit naming

ege0zcan: Actually it was a mistake on my part, they have different shapes I will update it

src/features/chat/types.ts
ege0zcan: I would argue for the opposite, there is no need to have anything more than a string for the quick action for now. ChatPrompt would add extra fields for no purpose

src/features/chat/index.ts
```diff
@@ -0,0 +1,41 @@
...
+    onCodeInsertToCursorPosition?: (handler: NotificationHandler<InsertToCursorPositionParams>) => void
+    onCopyCodeToClipboard?: (handler: NotificationHandler<CopyCodeToClipboardParams>) => void
+    onLinkClick?: (handler: NotificationHandler<LinkClickParams>) => void
+    onInfoLinkClick?: (handler: NotificationHandler<InfoLinkClickParams>) => void
+    onSourceLinkClick?: (handler: NotificationHandler<SourceLinkClickParams>) => void
+    // todo
+    onFollowUpClicked?: any
```
viktorsaws, line 40: Is this the only missing handler, can it be defined as part of this PR?

src/runtimes/webworker.ts
```diff
@@ -57,6 +60,12 @@ export const webworker = (props: RuntimeProps) => {
         },
     }
 
+    const chat: Chat = {
```
viktorsaws, line 63: This implementation of Chat interface is not complete. What's the plan for finishing it?

ege0zcan: Currently I would call it as a minimum viable product, it is sufficient for creating a simple chat server and we can continue iterating on it

viktorsaws (approved):

PR #67, opened 2024-03-07
removed recursive prop from readdir

mj632:
## Problem
readdir recursive only works if Node version is 20+. 
## Solution
Removing `recursive` optional prop and the param from `readdir` .
Here is a way to make readdir recursive - https://stackoverflow.com/a/45130990

## License


ege0zcan (approved):

mj632:
> Don't we control the nodejs version (because we ship it)? 

Yes, we do and the current version is 18. We need to update to V20  but need to check the roadmap for that before we upgrade so that we don't loose existing customers that are still on Node 18.

PR #88, opened 2024-04-12
[WIP] Enable injection of credentials provider

rtarcr:
DO NOT MERGE. THIS IS A WIP PR

## Problem
We want integrators to be able to inject their own credential provider. 
## Solution
Allow injecting the provider. [This line](https://github.com/aws/language-servers/blob/9dd93327a620f93967f1270d9f5548c85c7d8e3b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L362) needs to be modified to avoid calls when the method is not implemented.

## License


src/server-interface/auth.ts
```diff
@@ -9,5 +9,5 @@ export type Credentials = IamCredentials | BearerCredentials
 export interface CredentialsProvider {
     hasCredentials: (type: CredentialsType) => boolean
     getCredentials: (type: CredentialsType) => Credentials | undefined
-    getConnectionMetadata: () => ConnectionMetadata | undefined
+    getConnectionMetadata?: () => ConnectionMetadata | undefined
```
andredcoliveira, line 12: Since this is now optional, we should also verify whether it's defined before invoking it:
- https://github.com/aws/language-servers/blob/51d773407f98fc0088c925bf2b2514e818ce1f0d/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L362
- https://github.com/aws/language-servers/blob/51d773407f98fc0088c925bf2b2514e818ce1f0d/server/aws-lsp-codewhisperer/src/language-server/codeWhispererSecurityScanServer.ts#L46

Otherwise integrators need to add something like `getConnectionMetadata: () => undefined` to their `credentialsProvider` to prevent an unhandled error from crashing the server.

rtarcr: Indeed. I mentioned this in the PR description: 
> [This line](https://github.com/aws/language-servers/blob/9dd93327a620f93967f1270d9f5548c85c7d8e3b/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L362) needs to be modified to avoid calls when the method is not implemented.

Hope you didn't have to debug the server to find this out the hard way

andredcoliveira: My bad, I completely missed that. It's no problem, I figured it out in a couple of minutes :)

PR #91, opened 2024-04-16
Convert repo to monorepo and create new types package

rahmaniaam:
## Problem
We need chat types to be in a different "types" package

## Solution
This PR turns the repo into a monorepo. The original `@aws/language-server-runtimes` package is moved to live in `runtimes` folder in root.
The new types package `@aws/language-server-runtimes-types` lives in `types` folder, completely separate from the runtimes package which depends on this package. Structure should support future additions of types, not just chat types. 

## License


package.json
```diff
@@ -20,7 +24,7 @@
     },
     "scripts": {
         "clean": "rm -rf out",
-        "compile": "tsc --build",
+        "compile": "tsc --build && npm run compile --workspaces --if-present",
```
volodkevych, line 27: Given `types` package is inside of `language-server-runtimes` package, won't the package already contain `types` in it? Isn't it better to have `types` and `runtimes` as 2 fully separate packages?

rahmaniaam: The way I see it we want to eventually have a similar packaging structure like `vscode-languageserver`, where we can either install the whole `language-server-runtimes` or only select packages (`types`, `protocol`, `runtimes`, etc).

Making them 2 fully separate packages (basically separate repos?), I feel will introduce more delays in development

volodkevych: I do not suggest separate repository. Also, I'm not sure we'll split other folders in `language-server-runtimes` into packages. But I suggest to keep the 2 packages we are creating now (`types` and `runtime` containing all the rest) in the root of project so that they are separated. Currently, types package is inside the `runtime` package.

imykhai: afaik the root `package.json` remains private even when we want to export the whole package. It's just that there will be a separate subpackage that reexports the content from other sub packages. 
For example, here in [`vscodelanguageserver` repo the root package is private](https://github.com/microsoft/vscode-languageserver-node/blob/8e625564b531da607859b8cb982abb7cdb2fbe2e/package.json#L2)

imykhai (approved):

package.json
```diff
@@ -19,27 +23,15 @@
         "node": ">=18.0.0"
     },
     "scripts": {
-        "clean": "rm -rf out",
-        "compile": "tsc --build",
+        "clean": "ts-node ./script/clean.ts",
+        "compile": "tsc --build --verbose",
```
volodkevych, line 27: nit: verbose is probably leftover from debugging

volodkevych (approved):

PR #92, opened 2024-04-16
Allow UpdateCredentials request params by-name and by-position

imykhai:
## Problem
[JSON-RPC protocol](https://www.jsonrpc.org/specification) allows passing parameters as an array (positinal) or as an object (by names). For example, 
```
rpc call with positional parameters:
--> {"jsonrpc": "2.0", "method": "subtract", "params": [42, 23], "id": 1}
<-- {"jsonrpc": "2.0", "result": 19, "id": 1}

rpc call with named parameters:
--> {"jsonrpc": "2.0", "method": "subtract", "params": {"subtrahend": 23, "minuend": 42}, "id": 3}
<-- {"jsonrpc": "2.0", "result": 19, "id": 3}
```

Previously, the CodeWhisperer LSP Server allowed `'aws/credentials/token/update'` JSON-RPC requests to pass parameters by name and positional. 
But [a recent refactoring in `aws/language-server-runtimes` changed the implementation](https://github.com/aws/language-server-runtimes/commit/3c74fee7af9c0cb2792c5d9683bfd5afce1dfa7d#diff-8d9c26fb3fabe4ac9c52241e64600a45dbe451928e1201031acc4a0f61d12707R29) of this method:
```
export const bearerCredentialsUpdateRequestType = new ProtocolRequestType<
    UpdateCredentialsParams,
    null,
    void,
    void,
    void
>('aws/credentials/token/update')
```
Now it uses [`vscode-language-server-protocol`'s `ProtocolRequestType`](https://github.com/microsoft/vscode-languageserver-node/blob/main/protocol/src/common/messages.ts#L51) class which limits the requests to pass parameters by name only:
```

export class ProtocolRequestType<P, R, PR, E, RO> ... {
    //...
	public constructor(method: string) {
		super(method, ParameterStructures.byName);
	}
}
```

Thus when the VisualStudio sends the JSON-RPC message to pass the bearer token as positional arguments the server now responds with the error:
```
Request aws/credentials/token/update defines parameters by name but received parameters by position
```

The problem wasn't reproducible in the VSCode client because it sent the parameters by name. 

## Solution

A new `RequestType` is exported in protocol now - `AutoParameterStructuresProtocolRequestType`. The class repeats the `ProtocolRequestType` but it allows ParameterStructures both by-name and by-position. 

Updated `bearerCredentialsUpdateRequestType` and `iamCredentialsUpdateRequestType` to base on `AutoParameterStructuresProtocolRequestType`. 

Added a test case that checks the server allows the bearer token by-position as well as by-name.

## License

### DIFF
--- a/src/runtimes/auth/auth.test.ts
+++ b/src/runtimes/auth/auth.test.ts
@@ -4,7 +4,7 @@ import * as jose from 'jose'
 import { Duplex } from 'stream'
 import { Connection, createConnection } from 'vscode-languageserver/node'
 import { Auth } from './auth'
-import { UpdateCredentialsParams } from '../../protocol'
+import { ParameterStructures, UpdateCredentialsParams } from '../../protocol'
 import { IamCredentials, BearerCredentials, CredentialsType, CredentialsProvider } from '../../server-interface'
 
 export const credentialsProtocolMethodNames = {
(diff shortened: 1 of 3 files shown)


viktorsaws (approved):

src/runtimes/auth/auth.test.ts
```diff
@@ -113,6 +113,27 @@ describe('Auth', () => {
         assert(credentialsProvider.getCredentials('iam') === undefined)
     })
 
+    it('Handles Set Bearer credentials request when parameters sent by position', async () => {
```
rahmaniaam, line 116: Can we add test for by-name request as well? incase VSCode implementation changes as well

imykhai:
> nit: should we add documentation to refrain from using `ProtocolRequestType` from now?

@rahmaniaam , I think we need to further investigate why `vscode-language-server-node` made `by-name` the default behaviour before refraining from `ProtocolRequestType`

PR #95, opened 2024-04-22
Move types to types package

rahmaniaam:
## Problem
All type besides chat-types are still in the runtimes package

## Solution
- Move types to `types` package.
- Change re-export of  `vscode-languageserver-protocol` to `types` package in `runtimes` package
- Update dependencies

## License

### DIFF
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -1,7 +1,7 @@
 import { ParameterStructures, ProgressType, RegistrationType, RequestType } from 'vscode-languageserver-protocol'
 import { _EM } from 'vscode-jsonrpc'
 
-export { TextDocument, Position } from 'vscode-languageserver-textdocument'
+export * from '@aws/language-server-runtimes-types'
 
 // LSP protocol is a core dependency for LSP feature provided by runtimes.
 // Since we aim to provide whole range of LSP specification for Clients and Capabilities,
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -2,6 +2,7 @@ import { ParameterStructures, ProgressType, RegistrationType, RequestType } from
 import { _EM } from 'vscode-jsonrpc'
 
 export * from '@aws/language-server-runtimes-types'
+export { TextDocument } from 'vscode-languageserver-textdocument'
 
 // LSP protocol is a core dependency for LSP feature provided by runtimes.
 // Since we aim to provide whole range of LSP specification for Clients and Capabilities,
--- a/types/README.md
+++ b/types/README.md
@@ -1,0 +1,11 @@
+# Language Server Runtimes Types
+
+Language Server Runtimes Types is a package containing types used by the [Language Server Runtimes](../runtimes/) package.
+
+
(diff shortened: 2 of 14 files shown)


types/lsp.ts
```diff
@@ -0,0 +1,8 @@
+export { TextDocument, Position } from 'vscode-languageserver-textdocument'
```
viktorsaws, line 1: If we aim to have pure `types` package, we shouldn't re-export anything from vscode packages other than `vscode-languageserver-types`. 
'vscode-languageserver-textdocument' and 'vscode-languageserver-protocol' contain implementation of whole LSP protocol, which is re-exported here, but that defies the purpose of `types` package.

runtimes/protocol/lsp.ts
```diff
@@ -1,14 +1,7 @@
...
 import { _EM } from 'vscode-jsonrpc'
 
-export { TextDocument, Position } from 'vscode-languageserver-textdocument'
-
-// LSP protocol is a core dependency for LSP feature provided by runtimes.
-// Since we aim to provide whole range of LSP specification for Clients and Capabilities,
-// we re-exporting whole LSP protocol for usage.
-// Scoping it down is not practical due to large surface of protocol and types relationship.
-// It will be limiting implementors, if they choose to type their code with more specific types from LSP.
-export * from 'vscode-languageserver-protocol'
```
viktorsaws, line 11: I don't think this change in correct: this effectively moves whole LSP protocol to `types` package, which makes this package meaningless.

types/lsp.ts
rahmaniaam: Makes sense, we were importing many types directly from `vscode-languageserver-protocol`. Though I'm not sure about `vscode-languageserver-textdocument` as an implementation of LSP protocol, it seems to only define text document specific types which would fall under a "pure types" package criteria

rahmaniaam: vscode themselves seem to use `textdocument` to complement their `types` package https://github.com/microsoft/vscode-languageserver-node/blob/main/types/src/main.ts#L4517C1-L4518C1

types/README.md
```diff
@@ -0,0 +1,11 @@
+# Language Server Runtimes Types
+
+Language Server Runtimes Types is a package containing types used by the [Language Server Runtimes](../runtimes/) package.
+
+
+This package is independent of [VSCode Protocol](https://github.com/microsoft/vscode-languageserver-node/tree/main/protocol) type definitions.
```
viktorsaws, line 6: nit: would be better to make this documentation a bit more instructive, saying that types should not depend on implementation details from `protocol`.

types/lsp.ts
```diff
@@ -0,0 +1,3 @@
+export { TextDocument } from 'vscode-languageserver-textdocument'
```
viktorsaws, line 1: We should be able to use `export type` syntax here to signify that `TextDocument` is not supposed to be used downstream as an object with side-effect, but only provides a type.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html

viktorsaws (approved):

volodkevych (approved):

types/README.md
viktorsaws: (fixed inline)

types/lsp.ts
viktorsaws: (fixed)

PR #98, opened 2024-04-29
Add cursor state syncing to workspace

ege0zcan:
## Problem
LSP doesn't provide a way to sync cursor state from client to server. See some discussions here: https://github.com/microsoft/language-server-protocol/issues/718#issuecomment-1765870852

We want to provide the ability to sync this state. For example this could be used by chat servers to enrich the chat requests with the context of the currently highlighted code. 

## Solution
This change introduces a new notification type that will allow clients to send cursor state updates to servers. 

Clients will be free to determine how often they send the cursor update notification. They can send every cursor change, send the latest change every X seconds or send the notification only before sending a specific request (for example a code completion request) 

Tested the change on vscode using the test client in https://github.com/aws/language-servers 
Screenshot 2024-04-29 at 14 37 32

## License


runtimes/protocol/workspace.ts
```diff
@@ -0,0 +1,9 @@
+import { Range, Position, ProtocolNotificationType } from './lsp'
+
+export interface CursorStateUpdateParams {
+    cursorState: { position: Position } | { range: Range }
```
kmile, line 4: What about https://code.visualstudio.com/docs/editor/codebasics#_multiple-selections-multicursor ?

runtimes/protocol/workspace.ts
```diff
@@ -0,0 +1,9 @@
...
+}
+
+export const cursorStateUpdateNotificationType = new ProtocolNotificationType<CursorStateUpdateParams, void>(
```
kmile, line 7: Should this be part of `extensions` instead of `workspace`?

runtimes/server-interface/workspace.ts
```diff
@@ -27,4 +30,5 @@ export type Workspace = {
         isFile: (path: string) => Promise<boolean>
         remove: (dir: string) => Promise<void>
     }
+    onCursorStateUpdate: (handler: NotificationHandler<CursorStateUpdateParams>) => void
```
volodkevych, line 33: nit: `onCursorStateChange` fits a bit better for the name

volodkevych: If we included cursor position into chat & quick actions request, it would have been much less chatty.

ege0zcan: Yes, for now those two would be the only two places where we add the cursorstate param. But it can cause issues as the interfaces grow, we would need to add the cursor state as an optional parameter into many more requests

runtimes/protocol/workspace.ts
```diff
@@ -0,0 +1,9 @@
+import { Range, Position, ProtocolNotificationType } from './lsp'
+
+export interface CursorStateUpdateParams {
```
volodkevych, line 3: Can we move this to `types` package.

ege0zcan: Yes, still adjusting to the refactor

ege0zcan:
Closing this pull request, we decided not to move forward with this approach.

The chosen approach is here: https://github.com/aws/language-server-runtimes/pull/99

PR #101, opened 2024-04-30
Inject provider

rtarcr:
## Problem
For cases when credentials can not be sent over LSP, we want to be able to inject credentials into the runtime

## Solution
Enable injection of `credentialsProvider` as `RuntimeProps` 

## License


dogusata (approved):

runtimes/runtimes/runtime.ts
```diff
@@ -16,4 +16,8 @@ export type RuntimeProps = {
      * Name of the server used inside the runtime
      */
     name: string
+    /**
+     * Credential provider to overwrite the default Auth mechanism over LSP.
+     */
+    credentialsProvider?: CredentialsProvider
```
kmile, line 22: Is this a generic property that can be injected into any runtime? Or is this a specific property of specific runtimes?

For example: having a cookie-based credentials provider only makes sense in a `webworker` runtime. Having credentials loaded from environment variables only make sense from a `standalone` runtime. Other variants might exist. Could this be a property of specific runtimes instead?

rtarcr: Currently it is a property that can be used for any runtime that uses credentialsProvider (all of them as of now). 

With Auth over LSP, the client is responsible for pushing the right type of credentials and servers are being passed the same type of credentialsProvider regardless of what credentials are available. 

The same behavior applies here: the integrator is responsible for providing the right type of credentials and the runtime just passes it to the server. Whether integrators use cookies or env variables to get IAM credentials, both approaches would satisfy the interface from our point of view

rtarcr: Or do you mean that we should not allow integrators the ability to inject credentialsProvider but rather create different flavors of runtimes that have different Auth mechanisms?

PR #104, opened 2024-05-07
Create new subpackage for chat client ui types

ege0zcan:
## Description

Creating a new package for defining chat client ui contracts between `chat client <-> ide extension` and `ide extension <-> chat server` These types will be consumed by chat clients defined in `aws/language-servers` 

## License

### DIFF
--- a/chat-client-ui-types/package.json
+++ b/chat-client-ui-types/package.json
@@ -1,0 +1,23 @@
+{
+    "name": "@aws/chat-client-ui-types",
+    "version": "0.0.1",
+    "description": "Type definitions for Chat UIs in Language Servers and Runtimes for AWS",
+    "main": "index.js",
+    "scripts": {
+        "clean": "rm -rf out/",
+        "compile": "tsc --build",
+        "prepub:copyFiles": "cp ../.npmignore CHANGELOG.md ../LICENSE ../NOTICE README.md ../SECURITY.md package.json out/",
+        "prepub": "npm run clean && npm run compile && npm run prepub:copyFiles",
+        "pub": "cd out && npm publish"
+    },
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/aws/language-server-runtimes",
+        "directory": "chat-client-ui-types"
+    },
+    "author": "Amazon Web Services",
+    "license": "Apache-2.0",
+    "dependencies": {
+        "@aws/language-server-runtimes-types": "^0.0.2"
+    }
+}
--- a/chat-client-ui-types/uiContracts.ts
+++ b/chat-client-ui-types/uiContracts.ts
@@ -1,0 +1,102 @@
+import { ReferenceTrackerInformation, CodeSelectionType } from '@aws/language-server-runtimes-types'
+
+export type AuthFollowUpType = 'full-auth' | 're-auth' | 'missing_scopes' | 'use-supported-auth'
+export function isValidAuthFollowUpType(value: string): value is AuthFollowUpType {
+    return ['full-auth', 're-auth', 'missing_scopes', 'use-supported-auth'].includes(value)
+}
+
+export type GenericCommandVerb = 'Explain' | 'Refactor' | 'Fix' | 'Optimize'
+
+export const TAB_ID_RECEIVED = 'triggerTabIdReceived'
+export const SEND_TO_PROMPT = 'sendToPrompt'
+export const ERROR_MESSAGE = 'errorMessage'
+export const INSERT_TO_CURSOR_POSITION = 'insertToCursorPosition'
+export const AUTH_FOLLOW_UP_CLICKED = 'authFollowUpClicked'
+export const GENERIC_COMMAND = 'genericCommand'
+
+export type UiMessageCommand =
+    | typeof TAB_ID_RECEIVED
+    | typeof SEND_TO_PROMPT
+    | typeof ERROR_MESSAGE
+    | typeof INSERT_TO_CURSOR_POSITION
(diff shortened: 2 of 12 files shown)


chat-client-ui-types/serverContracts.ts
```diff
@@ -0,0 +1,28 @@
+import { TabAddParams, TabChangeParams, TabRemoveParams } from '@aws/language-server-runtimes-types'
+
+export const CHAT_PROMPT = 'aws/chat/sendChatPrompt'
+export const NEW_TAB_CREATED = 'aws/chat/tabAdd'
+export const TAB_REMOVED = 'aws/chat/tabChange'
+export const TAB_CHANGED = 'aws/chat/tabRemove'
```
volodkevych, line 6: The above two are not matching name to value.

chat-client-ui-types/uiContracts.ts
```diff
@@ -0,0 +1,104 @@
...
+    selection: string
+    eventId: string
+}
+
+export interface SendToPromptMessage {
+    command: typeof SEND_TO_PROMPT
+    params: SendToPromptParams
+}
+
+export interface InsertToCursorPositionParams {
```
volodkevych, line 54: Why is this type different from what we have in runtime types?

runtimes/server-interface/chat.ts
```diff
@@ -35,8 +33,6 @@ export type Chat = {
     onTabChange: (handler: NotificationHandler<TabChangeParams>) => void
     onTabRemove: (handler: NotificationHandler<TabRemoveParams>) => void
     onVote: (handler: NotificationHandler<VoteParams>) => void
```
volodkevych, line 35: Currently, it's based on telemetry. Do you think it can change in the future and voting becomes server concern?

ege0zcan: onVote can also be removed

runtimes/server-interface/chat.ts
```diff
@@ -35,8 +33,6 @@ export type Chat = {
...
-    onCodeInsertToCursorPosition: (handler: NotificationHandler<InsertToCursorPositionParams>) => void
```
volodkevych, line 38: Do you plan to reintroduce this later in the following PRs? I think `onCodeInsertToCursorPosition` will need to be sent to server at the end.

volodkevych (approved):

chat-client-ui-types/uiContracts.ts
kmile, line 1: Are these copied from a previous implementation?

ege0zcan: Some of these interfaces were temporarily defined here until a dedicated package was created-> https://github.com/aws/language-servers/blob/276b7bcac150bc5cb6011544c0d52de81f779e48/chat-client/src/contracts/uiContracts.ts

The rest of the interfaces are new

kmile: Ok, they are not pulled in from e.g., https://github.com/aws/aws-toolkit-vscode/blob/20c948be270c04d889e865bda84dc6b89ecfed25/packages/core/src/amazonq/webview/ui/connector.ts?

chat-client-ui-types/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "name": "@aws/chat-client-ui-types",
+    "version": "0.0.1",
+    "description": "Type definitions for Chat UIs in Language Servers and Runtimes for AWS",
+    "main": "index.js",
+    "scripts": {
+        "clean": "rm -rf out/",
+        "compile": "tsc --build",
+        "prepub:copyFiles": "cp ../.npmignore CHANGELOG.md ../LICENSE ../NOTICE README.md ../SECURITY.md package.json out/",
+        "prepub": "npm run clean && npm run compile && npm run prepub:copyFiles",
+        "pub": "cd out && npm publish"
```
viktorsaws, line 11: nit: for simple npm package (not monorepo), it's enough and more strainghtforward to publish from root of package, instead of `/out` directory. Not critical, but would be nice to simplify.

We had to publish `runtimes` from `out` for flatten import paths for multiple subfolders. Downside of this is that `npm link` has to be performed from `/out` dir as well during development, which is a bit more confusing.

chat-client-ui-types/uiContracts.ts
```diff
@@ -0,0 +1,92 @@
+import { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
+export { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
+
+export type AuthFollowUpType = 'full-auth' | 're-auth' | 'missing_scopes' | 'use-supported-auth'
+export function isValidAuthFollowUpType(value: string): value is AuthFollowUpType {
```
viktorsaws, line 5: Why are we exporting functions from ` types` package? Ideally `types` package is simply types without any side-effects and implementations.

ege0zcan: Ideally yes, but there is a strict dependency from ui types on the runtime types. The alternative is to define the same interface here which could cause issues in the future

viktorsaws: I was referring to `isValidAuthFollowUpType` function, but I guess it's fine to provide it as a helper type guard for consuming projects.

chat-client-ui-types/package.json
ege0zcan: I see, that makes sense. I will create new PR to update this

ege0zcan: https://github.com/aws/language-server-runtimes/pull/114

PR #115, opened 2024-05-16
Telemetry endpoint in protocol

rahmaniaam:
## Problem
Telemetry endpoint in the protocol is not yet exposed

## Solution
Enable endpoint in telemetry feature to enable clients sending telemetry to server

Also added additional UI chat contracts for copyCodeToClipboard and vote events 

This PR supports implementation in https://github.com/aws/language-servers/pull/249

## License

### DIFF
--- a/chat-client-ui-types/src/uiContracts.ts
+++ b/chat-client-ui-types/src/uiContracts.ts
@@ -1,22 +1,9 @@
-import {
-    CodeSelectionType,
-    InsertToCursorPositionParams,
-    ReferenceTrackerInformation,
-} from '@aws/language-server-runtimes-types'
-export {
-    CodeSelectionType,
-    InsertToCursorPositionParams,
-    ReferenceTrackerInformation,
-} from '@aws/language-server-runtimes-types'
+import { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
+export { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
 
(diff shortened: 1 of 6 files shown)


chat-client-ui-types/src/uiContracts.ts
```diff
@@ -67,6 +80,24 @@ export interface AuthFollowUpClickedMessage {
     params: AuthFollowUpClickedParams
 }
 
+export interface CopyCodeToClipboardParams {
```
volodkevych, line 83: Why do we need those in UI contracts? Will extension need to know the telemetry event to publish it to DWH?

rahmaniaam: It's not necessary, but I think it's better to have a format that the client conforms to and possible future requirements where the data are expected

chat-client-ui-types/src/uiContracts.ts
```diff
@@ -67,6 +80,24 @@ export interface AuthFollowUpClickedMessage {
...
+    tabId: string
+    messageId: string
+    code?: string
+    type?: CodeSelectionType
+    referenceTrackerInformation?: ReferenceTrackerInformation[]
+    eventId: string
```
ege0zcan, line 89: is this not optional?

chat-client-ui-types/src/uiContracts.ts
rahmaniaam: It's more for the chat-client to conform it's data before sending to Toolkit, updated to reflect this comment in the adjacent PR https://github.com/aws/language-servers/pull/249#discussion_r1603316608

volodkevych (approved):

PR #121, opened 2024-05-29
Add missing field in ChatItemAction and adjust link params fields

rahmaniaam:
## Problem
ChatItemAction is missing optional `type` field (needed for auth follow-ups). 
FeedbackParams is missing optional `eventId` field
InfoLinkClick does not pass `messageId` but is required in it's params

## Solution
Add missing fields and rearrange params for links interface structure

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -6,6 +6,7 @@ export interface ChatItemAction {
     prompt?: string
     disabled?: boolean
     description?: string
+    type?: string
 }
 
 export interface SourceLink {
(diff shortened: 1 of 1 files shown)


types/chat.ts
```diff
@@ -109,15 +110,19 @@ export interface InsertToCursorPositionParams {
     totalCodeBlocks?: number
 }
 
-export interface LinkClickParams {
+export interface InfoLinkClickParams {
     tabId: string
-    messageId: string
     link: string
     mouseEvent?: MouseEvent
```
ege0zcan, line 116: I suggest we remove mouseevent completely at this point while we are making changes

ege0zcan (approved):

PR #127, opened 2024-06-03
Upgrade chat-client-ui-types version to 0.1.0

rahmaniaam:
## Problem
We want to use the latest chat-client-ui-types
## Solution
Release a new version

## License

### DIFF
--- a/chat-client-ui-types/package.json
+++ b/chat-client-ui-types/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@aws/chat-client-ui-types",
-    "version": "0.0.3",
+    "version": "0.1.0",
     "description": "Type definitions for Chat UIs in Language Servers and Runtimes for AWS",
     "main": "/out/index.js",
     "scripts": {
--- a/chat-client-ui-types/CHANGELOG.md
+++ b/chat-client-ui-types/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [0.1.0] - 2024-06-03
+
+- Update package folder structure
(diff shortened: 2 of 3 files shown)


chat-client-ui-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@aws/chat-client-ui-types",
-    "version": "0.0.3",
+    "version": "0.1.0",
```
ege0zcan, line 3: Why are we doing a new minor version rather than a new patch version? 

The changed folder structure shouldn't change how the package is consumed

rahmaniaam: Because now the `out/` package is being published as well, the imports have to be updated to `@aws/chat-client-ui-types/out`

ege0zcan: The out package is being published but consumers won't have to update their imports because the entry point in package.json points to `out/index.js`  https://github.com/aws/language-server-runtimes/blob/main/chat-client-ui-types/package.json#L5 so imports from `@aws/chat-client-ui-types` should work as before

ege0zcan: Though, this is misconfigured right now: https://github.com/aws/language-server-runtimes/blob/main/chat-client-ui-types/package.json#L5

We need to change `/out/index.js` to `out/index.js`

rahmaniaam: Imports did not work because of typo in main entry, fixing it in https://github.com/aws/language-server-runtimes/pull/128

I'll close this PR in favour of a new one

PR #131, opened 2024-06-04
Added custom awsServerCapabilities registration for servers at initialization flow

viktorsaws:
## Problem
We want to allow server implementations to register it's supported quick actions for Chat feature, to properly render chat at start time.

## Solution

Extending default LSP `InitializeResult` objects with new backwards-compatible optional `awsServerCapabilities` attribute, scoped with `AWS`. This attribute is used to allow Server implementations to register customer features, defined by Language Servers runtimes protocol. First feature which is added is `chatQuickActionsProvider` for registering Quick Actions in Chat Client implementation (https://github.com/aws/language-servers/tree/main/chat-client).

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #134, opened 2024-06-05
Extend server handshake to register chat quick actions

ege0zcan:
Following up on: https://github.com/aws/language-server-runtimes/pull/131, this PR takes only the chat quick action related changes from the other PR. Copying the description from the other PR:

## Problem
We want to allow server implementations to register it's supported quick actions for Chat feature, to properly render chat at start time.

## Solution

Extending default LSP `InitializeResult` objects with new backwards-compatible optional `awsServerCapabilities` attribute, scoped with `AWS`. This attribute is used to allow Server implementations to register customer features, defined by Language Servers runtimes protocol. First feature which is added is `chatQuickActionsProvider` for registering Quick Actions in Chat Client implementation (https://github.com/aws/language-servers/tree/main/chat-client).

## License

### DIFF
--- a/runtimes/protocol/chat.ts
+++ b/runtimes/protocol/chat.ts
@@ -31,6 +31,34 @@ import {
     TAB_REMOVE_NOTIFICATION_METHOD,
 } from './lsp'
 
+/**
+ * Configuration object for chat quick action.
+ */
+export interface QuickActionCommand {
+    command: string
+    disabled?: boolean
+    description?: string
+    placeholder?: string
+}
+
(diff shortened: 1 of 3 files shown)


viktorsaws:
What I didn't do in my PR was to document expectation for initialise handshake and how it extends default LSP one. We will need to properly document all protocol updates to clearly specify all extensions we are making and which UX and features they enable, and what are expectation from clients to enable those features. (It's not a blocker for merging this change, but something we need to address in the future).

PR #140, opened 2024-06-13
Move QuickActionsOptions to runtimes-types and consume it in chat-ui-types

ege0zcan:
## Problem
QuickActionsOptions needs to be used by the chat-client (see https://github.com/aws/language-servers/pull/302 ) for setting the available quick action options dynamically

## Solution
Move the type out of protocol and into runtimes-types and also import and re-export it from the chat-ui-types  

## License


chat-client-ui-types/src/uiContracts.ts
```diff
@@ -1,5 +1,5 @@
-import { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
-export { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
+import { InsertToCursorPositionParams, QuickActionsOptions } from '@aws/language-server-runtimes-types'
+export { InsertToCursorPositionParams, QuickActionsOptions } from '@aws/language-server-runtimes-types'
```
volodkevych, line 2: Why does client/destination need to know about these options? Is Toolkit expected to manipulate the options somehow instead of just forwarding the initialize response?

chat-client-ui-types/src/uiContracts.ts
```diff
@@ -32,6 +34,8 @@ export type UiMessageParams =
     | AuthFollowUpClickedParams
     | GenericCommandParams
     | ErrorParams
+    | SendToPromptParams
+    | QuickActionsOptions
```
volodkevych, line 38: Shouldn't chat client receive the whole `initializeResult` from server?

viktorsaws: Whole `initializeResult` can contain too many not relevant data. 

It maybe better to specify more generic wrapping initialise object expected by chat client instead of `QuickActionsOptions` specifically. This way this object can be extended. What happens if tomorrow we need to extend this to more data?

chat-client-ui-types/src/uiContracts.ts
ege0zcan: I agree with Viktor that passing the whole InitializeResult would be too wasteful for what we are trying to achieve. I can create a more generic type like `defaultTabDataOptions` that only contains quick actions for the time being and that can be extended in the future, but I don't think initializeResult is the right approach

viktorsaws: `defaultTabData` prefix as a public-facing interface is a bit confusing outside of chat client codebase imo. Something more concrete would be better, like `chatOptions` or `chatConfiguration`?

volodkevych: I'm fine with the suggested approach.
But the original question is still in place: why does this options type need to be exposed to client/destination?

viktorsaws (approved):

PR #143, opened 2024-06-15
Provider fqn worker pool as a feature

makenneth:
## Description
This approach gets rid of the need to publish since the environment split is at the server bundle level. I think this was the eventual goal of the fqn feature so we might as well if there is an ongoing issue of getting the fqn feature released.

The logic is a more polished version of https://github.com/aws/language-servers/tree/main/core/aws-lsp-fqn. The extract function is now cancellable and supports extractNames (without extent)

We have to make build configuration clear to any consumers though https://github.com/aws/language-servers/pull/298 (for node)

## License


runtimes/runtimes/fqn/common/fqnExtractor.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        )
+    })
+
+    it('calls the corresponding fqn function', async () => {
+        let result = await findNames(fqn, {
+            languageId: 'typescript',
+            fileText: mockFileText,
+        })
+
+        sinon.assert.calledOnceWithMatch(typescriptStub, mockFileText)
```
justinmk3, line 48: why are any of these tests using mocks (sinon)? can't they open an actual text document and check the actual fqn results? there is no network/service calls involved, right?

runtimes/runtimes/fqn/common/commonFqnWorkerPool.ts
```diff
@@ -0,0 +1,59 @@
...
+import {
+    Cancellable,
+    ExtractFqnInput,
+    ExtractFqnResult,
+    IFqnWorkerPool,
+    Logging,
+    WorkerPoolConfig,
+} from '../../../server-interface'
+
+export class CommonFqnWorkerPool implements IFqnWorkerPool {
```
justinmk3, line 12: why is a worker pool needed? it might be obvious now, but others working on this project in the future won't have the context that you have

runtimes/runtimes/fqn/node/fqnWorkerPool.ts
```diff
@@ -0,0 +1,9 @@
+import path = require('path')
+import { CommonFqnWorkerPool } from '../common/commonFqnWorkerPool'
+import { WorkerPoolConfig } from '../../../server-interface'
+
+export class FqnWorkerPool extends CommonFqnWorkerPool {
```
justinmk3, line 5: why is a worker pool needed? it might be obvious now, but others working on this project in the future won't have the context that you have 

a brief docstring on this class can give insight.

PR #145, opened 2024-06-19
Add encryption to chat

ege0zcan:
## Description
This PR introduces the encrypted chat class which ensures that quick action and chat prompt communication passing through the LSP connection is encrypted. The runtime handles the encryption and decryption and servers don't need to change anything

## License

### DIFF
--- a/runtimes/protocol/chat.ts
+++ b/runtimes/protocol/chat.ts
@@ -15,7 +15,6 @@ import {
     TabRemoveParams,
     ProtocolNotificationType,
     ProtocolRequestType,
-    ProgressToken,
     CHAT_REQUEST_METHOD,
     END_CHAT_REQUEST_METHOD,
     FEEDBACK_NOTIFICATION_METHOD,
@@ -29,6 +28,10 @@ import {
     TAB_ADD_NOTIFICATION_METHOD,
     TAB_CHANGE_NOTIFICATION_METHOD,
     TAB_REMOVE_NOTIFICATION_METHOD,
+    AutoParameterStructuresProtocolRequestType,
+    EncryptedChatParams,
+    EncryptedQuickActionParams,
+    QuickActionResult,
 } from './lsp'
 
 /**
@@ -59,19 +62,23 @@ export interface QuickActionsOptions {
     quickActionsCommandGroups: QuickActionCommandGroup[]
 }
 
-export interface ChatRequest extends ChatParams {
-    partialResultToken?: ProgressToken
-}
-
-export const chatRequestType = new ProtocolRequestType<ChatRequest, ChatResult, ChatResult, void, void>(
-    CHAT_REQUEST_METHOD
-)
+export const chatRequestType = new AutoParameterStructuresProtocolRequestType<
+    ChatParams | EncryptedChatParams,
+    ChatResult | string,
+    ChatResult | string,
+    void,
+    void
+>(CHAT_REQUEST_METHOD)
 export const endChatRequestType = new ProtocolRequestType<EndChatParams, EndChatResult, never, void, void>(
     END_CHAT_REQUEST_METHOD
 )
-export const quickActionRequestType = new ProtocolRequestType<QuickActionParams, ChatResult, ChatResult, void, void>(
-    QUICK_ACTION_REQUEST_METHOD
-)
+export const quickActionRequestType = new AutoParameterStructuresProtocolRequestType<
+    QuickActionParams | EncryptedQuickActionParams,
+    QuickActionResult | string,
+    QuickActionResult | string,
+    void,
+    void
+>(QUICK_ACTION_REQUEST_METHOD)
 export const readyNotificationType = new ProtocolNotificationType<void, void>(READY_NOTIFICATION_METHOD)
 export const feedbackNotificationType = new ProtocolNotificationType<FeedbackParams, void>(FEEDBACK_NOTIFICATION_METHOD)
 export const tabAddNotificationType = new ProtocolNotificationType<TabAddParams, void>(TAB_ADD_NOTIFICATION_METHOD)
--- a/runtimes/runtimes/chat/encryptedChat.ts
+++ b/runtimes/runtimes/chat/encryptedChat.ts
@@ -51,29 +51,32 @@ export class EncryptedChat implements Chat {
     }
 
     public onChatPrompt(handler: RequestHandler<ChatParams, ChatResult | null | undefined, ChatResult>) {
-        this.connection.onRequest(chatRequestType, async (request: ChatParams | EncryptedChatParams) => {
-            request = request as EncryptedChatParams
-            // decrypt the request params
-            let decryptedRequest = (await this.decodeRequest(request)) as ChatParams
+        this.connection.onRequest(
+            chatRequestType,
+            async (request: ChatParams | EncryptedChatParams, cancellationToken: CancellationToken) => {
+                request = request as EncryptedChatParams
+                // decrypt the request params
+                let decryptedRequest = (await this.decodeRequest(request)) as ChatParams
 
-            // make sure we don't lose the partial result token
-            if (request.partialResultToken) {
-                decryptedRequest.partialResultToken = request.partialResultToken
+                // make sure we don't lose the partial result token
+                if (request.partialResultToken) {
+                    decryptedRequest.partialResultToken = request.partialResultToken
+                }
+
+                // call the handler with plaintext
+                const response = (await handler(decryptedRequest, cancellationToken)) as ChatResult
+                // encrypt the response
+                const encryptedResponse = await this.encryptObject(response as JWTPayload)
+                // send it back
+                return encryptedResponse
             }
-
-            // call the handler with plaintext
-            const response = (await handler(decryptedRequest, CancellationToken.None)) as ChatResult
-            // encrypt the response
-            const encryptedResponse = await this.encryptObject(response as JWTPayload)
-            // send it back
-            return encryptedResponse
-        })
+        )
     }
 
     public onQuickAction(handler: RequestHandler<QuickActionParams, QuickActionResult, void>) {
         this.connection.onRequest(
             quickActionRequestType,
-            async (request: QuickActionParams | EncryptedQuickActionParams) => {
+            async (request: QuickActionParams | EncryptedQuickActionParams, cancellationToken: CancellationToken) => {
                 request = request as EncryptedQuickActionParams
                 // decrypt the request params
                 let decryptedRequest = (await this.decodeRequest(request)) as QuickActionParams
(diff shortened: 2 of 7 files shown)


runtimes/runtimes/standalone.ts
```diff
@@ -255,3 +269,9 @@ export const standalone = (props: RuntimeProps) => {
         lspConnection.listen()
     }
 }
+
+async function encryptRequest(request: Object, key: string): Promise<string> {
```
viktorsaws, line 273: nit: maybe better to move this function to separate util file, or to https://github.com/aws/language-server-runtimes/blob/main/runtimes/runtimes/auth/standalone/encryption.ts to keep helper functions related to encryption in single place

runtimes/runtimes/standalone.ts
```diff
@@ -102,6 +106,26 @@ export const standalone = (props: RuntimeProps) => {
         } else {
             lspConnection.console.info('Runtime: Initializing runtime without encryption')
             auth = new Auth(lspConnection)
+
+            chat = {
```
viktorsaws, line 110: Is moving initialisation of chat object out of `initializeRuntime` necessary? It still can be done inside in the condition

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -0,0 +1,156 @@
...
+    endChatRequestType,
+    CancellationToken,
+    EncryptedChatParams,
+    EncryptedQuickActionParams,
+    quickActionRequestType,
+} from '../../protocol'
+import { Chat } from '../../server-interface'
+import { CredentialsEncoding } from '../auth/standalone/encryption'
+
+export class EncryptedChat implements Chat {
```
viktorsaws, line 40: Most of methods implemented here are duplicating non-encrypted version of Chat in the runtime. It's better to convert non-encrypted object to new class and move duplicated methods to shared parent abstract class. This way encryption-specific implementations classes will be much leaner

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -0,0 +1,156 @@
...
+            const response = (await handler(decryptedRequest, CancellationToken.None)) as ChatResult
+            // encrypt the response
+            const encryptedResponse = await this.encryptRequest(response as JWTPayload)
+            // send it back
+            return encryptedResponse
+        })
+    }
+
+    public onQuickAction(handler: RequestHandler<QuickActionParams, QuickActionResult, void>) {
+        this.connection.onRequest(
```
viktorsaws, line 74: nit: code duplication in handling encrypted chatPrompt and quickAction could be reduced

runtimes/runtimes/standalone.ts
```diff
@@ -232,6 +239,13 @@ export const standalone = (props: RuntimeProps) => {
                 sendProgress: <P>(type: ProgressType<P>, token: ProgressToken, value: P) => {
                     return lspConnection.sendProgress(type, token, value)
                 },
+                sendEncryptedProgress: async <P>(type: ProgressType<P>, token: ProgressToken, value: P) => {
```
viktorsaws, line 242: Exposing `sendEncryptedProgress` to server implementations means that servers that want to use encryption must use this method instead of plain `sendProgress` in their business logic. 
I think it's not correct, and the condition implemented in this method should be moved to `sendProgress` method itself. This way, encrypted mode will be triggered under the hood, depending if encryption mode was enabled at server start time by client.

runtimes/protocol/chat.ts
```diff
@@ -59,19 +62,23 @@ export interface QuickActionsOptions {
...
+    void
+>(CHAT_REQUEST_METHOD)
 export const endChatRequestType = new ProtocolRequestType<EndChatParams, EndChatResult, never, void, void>(
     END_CHAT_REQUEST_METHOD
 )
-export const quickActionRequestType = new ProtocolRequestType<QuickActionParams, ChatResult, ChatResult, void, void>(
-    QUICK_ACTION_REQUEST_METHOD
-)
+export const quickActionRequestType = new AutoParameterStructuresProtocolRequestType<
+    QuickActionParams | EncryptedQuickActionParams,
```
kmile, line 76: Do we have to change every chat request type to use these encrypted parameter versions or is this something we can solve at the transport layer?

Are these `Encrypted*` exposed to servers implementors?

runtimes/runtimes/chat/encryptedChat.ts
kmile: Do we even need strong typing of the encrypted versions? Can we operate encryption at the lower level `params` and `results` members of the `json-rpc` requests/responses/notifications?

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -0,0 +1,156 @@
...
+                // send it back
+                return encryptedResponse
+            }
+        )
+    }
+    public onEndChat(handler: RequestHandler<EndChatParams, boolean, void>) {
+        return this.connection.onRequest(endChatRequestType, handler)
+    }
+
+    public onSendFeedback(handler: NotificationHandler<FeedbackParams>) {
```
kmile, line 99: Why don't we encrypt this or any other of the chat messages?

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -0,0 +1,156 @@
...
+
+            // make sure we don't lose the partial result token
+            if (request.partialResultToken) {
+                decryptedRequest.partialResultToken = request.partialResultToken
+            }
+
+            // call the handler with plaintext
+            const response = (await handler(decryptedRequest, CancellationToken.None)) as ChatResult
+            // encrypt the response
+            const encryptedResponse = await this.encryptRequest(response as JWTPayload)
```
kmile, line 67: ` encryptRequest`? Shouldn't that be `encryptResponse` or `encryptObject`?

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -0,0 +1,156 @@
...
+        private readonly connection: Connection,
+        key: string,
+        encoding?: CredentialsEncoding
+    ) {
+        this.key = Buffer.from(key, 'base64')
+        this.encoding = encoding
+    }
+
+    public onChatPrompt(handler: RequestHandler<ChatParams, ChatResult | null | undefined, ChatResult>) {
+        this.connection.onRequest(chatRequestType, async (request: ChatParams | EncryptedChatParams) => {
```
kmile, line 54: Does this have to be `ChatParams | EncryptedChatParams`? At this point we're only dealing in `EncryptedChatParams`, right?

runtimes/protocol/chat.ts
```diff
@@ -59,19 +62,23 @@ export interface QuickActionsOptions {
...
 }
 
-export interface ChatRequest extends ChatParams {
-    partialResultToken?: ProgressToken
-}
-
-export const chatRequestType = new ProtocolRequestType<ChatRequest, ChatResult, ChatResult, void, void>(
-    CHAT_REQUEST_METHOD
-)
+export const chatRequestType = new AutoParameterStructuresProtocolRequestType<
```
kmile, line 65: Is this necessary? Can we map to `EncyptedChatParams` inside the `EncryptedChat`  class without defining a union type on the request?

runtimes/runtimes/chat/encryptedChat.ts
kmile: I know that this `chatRequestType` defines the union type, but could we simply not do that here but define only an `encryptedChatRequestType`?

runtimes/runtimes/chat/encryptedChat.ts
kmile: I feel this whole class can be replaced with a generated type and a generic implementation of `onRequest` that automatically converts the intended `*RequestType` to the corresponding `Encrypted*RequestType` without typing it out explicitly.

kmile:
This could definitely benefit from having unit tests and explicit definition of when encryption is desired and/or required.

What does the client side look like?

runtimes/runtimes/chat/encryptedChat.ts
ege0zcan: We are only dealing with EncryptedParams but the signature for `chatRequestType` expect both of these types, otherwise typescript complains

runtimes/protocol/chat.ts
ege0zcan: I think you are right, let me check

ege0zcan:
> This could definitely benefit from having unit tests and explicit definition of when encryption is desired and/or required.
> 
> What does the client side look like?

Tests are definitely next on my list as a fast followup, example client looks like the following: https://github.com/aws/language-servers/pull/319/files

runtimes/runtimes/standalone.ts
ege0zcan: Not all progress should be encrypted though, only progress for specific types of requests need to be encrypted. 

Right now either approach is not ideal IMO, in the approach that I am proposing we expect the server to be aware that it is sending sensitive progress and explicitly use the encrypted progress. This ensures that only responses that need encryption get encrypted but this also means server implementor has to know about the detail of it


In the alternative approach you are proposing, everything will be encrypted and client would need to decrypt everything. The advantage is that server doesn't need to know or do anything extra

runtimes/runtimes/chat/encryptedChat.ts
ege0zcan: I spent some time trying to use generics and define a single function but it was taking long as it was not as straightforward as I expected so I left that for later

runtimes/runtimes/standalone.ts
viktorsaws: > This ensures that only responses that need encryption get encrypted but this also means server implementor has to know about the detail of it

I'm afraid that this detail pushes too much responsibility to individual Capability server implementation. How the server implementation will decide when to send encrypted vs. regular token? Capability server will need to know in which mode server was started if encryption was used or not.
What will happen if Capability calls `sendEncryptedProgress`, but server was started without encryption? The message will be dropped as I read it now.

Why do we need to support encrypted version of `sendProgress` at all? Why can't client send in unencrypted?

viktorsaws (approved):

kmile (approved):

ege0zcan:
Talked with @kmile and @viktorsaws offline. We all agree that there is room for improvement about these changes, I will merge this PR and work on these followups 

Major:
- This approach is limited to encryption on only 2 request types, that means clients have to know about these 2 request types and ensure proper encryption and decryption for these requests. In the future if we want to encrypt more requests, that would be a backwards incompatible breaking change for clients. So we need to think of a mechanism that would allow the runtime to understand if a request is encrypted or unencrypted and accordingly, similarly clients also need a similar mechanism for when they receive a response. If a client is hardcoded to only decrypt 2 requests, if in the future another request type starts being encrypted, the client would not try to decrypt that and it would cause a problem 

Minor:
- Add unit tests
- Clean up the code, minimize duplication

PR #169, opened 2024-07-15
Fix InitilizeResult merge logic

rahmaniaam:
## Problem
Servers couldn't run when multiple executeCommandProviders are given

## Solution
Add conditional to InitializeResult merge logic when values are arrays

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/util.ts
+++ b/runtimes/runtimes/lsp/router/util.ts
@@ -25,11 +25,13 @@ export function asPromise(value: any): Promise<any> {
 }
 
 export function mergeObjects(obj1: any, obj2: any) {
-    const merged: any = {}
+    let merged: any = {}
 
     for (let key in obj1) {
         if (obj1.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


runtimes/runtimes/lsp/router/util.ts
```diff
@@ -25,11 +25,13 @@ export function asPromise(value: any): Promise<any> {
 }
 
 export function mergeObjects(obj1: any, obj2: any) {
```
imykhai, line 27: seems like maintaining this function can be brittle. 
Might be worth considering replacing it with lodash's merge - https://lodash.com/docs/4.17.15#merge

https://www.npmjs.com/package/lodash

imykhai (approved):

runtimes/runtimes/lsp/router/util.ts
rahmaniaam: interesting, skimming the details it does appear to work for mixed data structs and does so recursively

PR #185, opened 2024-08-15
Add window showMessage and showDocument LSP features to runtimes

viktorsaws:
## Problem
Runtimes don't support Window showMessage features.

## Solution

Added support and bindings for 3 Window LSP features:
* `window/showMessage` notification ([spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage))
* `window/showMessageRequest` request ([spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest))
* `window/showDocument` request ([spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument))

Adding bindings for these features allows Capability servers to request displaying UI flows in host clients natively. This allows to implement UX flows. whenever required by Capability business logic.

## License

### DIFF
--- a/runtimes/server-interface/lsp.ts
+++ b/runtimes/server-interface/lsp.ts
@@ -35,6 +35,8 @@ import {
     ShowMessageParams,
     ShowMessageRequestParams,
     MessageActionItem,
+    ShowDocumentParams,
+    ShowDocumentResult,
 } from '../protocol'
 
 // Re-export whole surface of LSP protocol used in Runtimes.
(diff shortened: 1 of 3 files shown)


runtimes/server-interface/lsp.ts
```diff
@@ -94,6 +99,11 @@ export type Lsp = {
         getConfiguration: (section: string) => Promise<any>
         onDidChangeWorkspaceFolders: (handler: NotificationHandler<DidChangeWorkspaceFoldersParams>) => void
     }
+    window: {
+        showMessage: (params: ShowMessageParams) => Promise<void>
```
volodkevych, line 103: You seem to be using only `showMessage` for now in the related PR. Do you plan to use the rest in the following PRs?

viktorsaws: I didn't plan to use them, but I don't see a reason why not to add them now, since I'm touching this feature. 
If there is a need in other 2 tomorrow, they will be readily available.

PR #193, opened 2024-08-26
Accept Client metadata with LSP Initialization and inject standard UserAgent to Server implementations from the Runtime

viktorsaws:
## Problem

Currently Server implementations don't have access to actual information about client host environment, which can be used to construct detailed metadata. Such data can be used in cases, when connected clients and host environments need to be unambiguously identified, for example to attach granular UserAgent header for HTTP calls or AWS SDK calls done by the server.

## Solution
* Extended Language Server Protocol `InitializeParams` sent from LSP Client to the Runtime. This PR proposes to use unspecified standard `initializationOptions` field, and extend it with custom `aws` property, which isolates data for AWS Runtimes.
* Extended `InitializeParams` params passed from the Runtime to Server implementation, defined in `server-interface/lsp` interface.
* When Server's initialize method is called at startup, new `awsRuntimeMetadata` field is added, with interface specific for Runtimes. `customUserAgent` value is computed using data passed in new `initializationOptions.aws` and passed down each Server at initialization.
* Since userAgent field is highly specific for each language server implementation, we leave computation of the value based of available fields up to Server implementation

## License

### DIFF
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -46,23 +46,10 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
  */
 export interface AWSClientInitializationOptions {
     /**
-     * Custom client branding information.
+     * Custom UserAgent suffix value, computed by client and passed down to Server implementations at startup.
+     * The value can be used by Servers to setup custom user agent headers for HTTP or AWS SDK calls.
      */
-    product?: {
-        name: string
-        version: string
-    }
-    /**
-     * Client host platform information.
-     */
-    platform?: {
-        name: string
-        version: string
-    }
-    /**
-     * Custom Client ID value, set based on client extension internal business logic.
-     */
-    clientId?: string
+    customUserAgentSuffix?: string
 }
 
 /**
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -46,10 +46,29 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
  */
 export interface AWSClientInitializationOptions {
     /**
-     * Custom UserAgent suffix value, computed by client and passed down to Server implementations at startup.
-     * The value can be used by Servers to setup custom user agent headers for HTTP or AWS SDK calls.
+     * Custom client branding information.
      */
-    customUserAgentSuffix?: string
+    product?: {
+        /**
+         * Product name, may contain spaces.
+         */
+        name: string
+        version: string
+    }
+    /**
+     * Client host platform information.
+     */
+    platform?: {
+        /**
+         * Platform name, may contain spaces.
+         */
+        name: string
+        version: string
+    }
+    /**
+     * Custom Client ID value, set based on client extension internal business logic.
+     */
+    clientId?: string
 }
 
 /**
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -1,5 +1,6 @@
 import { _EM } from 'vscode-jsonrpc'
 import {
+    InitializeParams as _InitializeParamsBase,
     InitializeResult as InitializeResultBase,
     ParameterStructures,
     ProgressType,
@@ -37,6 +38,38 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
 
     public constructor(method: string) {
         super(method, ParameterStructures.auto)
+    }
+}
+
+/**
+ * Extra Client information passed during initialization handshake from Client to Server.
+ */
+export interface AWSClientInitializationOptions {
+    /**
+     * Custom client branding information.
+     */
+    product?: {
+        name: string
+        version: string
+    }
+    /**
+     * Custom client host platform information.
+     */
+    platform?: {
+        name: string
+        version: string
+    }
+    clientId?: string
+}
+
+/**
+ * Extended AWS Runtimes InitializeParams interface,
+ * sent from Client to Server as part of [`initialize`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) request
+ */
+export interface InitializeParams extends _InitializeParamsBase {
+    initializationOptions?: {
+        [key: string]: any
+        aws: AWSClientInitializationOptions
     }
 }
 
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -1,7 +1,7 @@
 import { _EM } from 'vscode-jsonrpc'
 import {
     InitializeParams as _InitializeParamsBase,
-    InitializeResult as InitializeResultBase,
+    InitializeResult as _InitializeResultBase,
     ParameterStructures,
     ProgressType,
     RegistrationType,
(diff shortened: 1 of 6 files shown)


runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,41 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
 
+/**
+ * Extra Client information passed during initialization handshake from Client to Server.
+ */
+export interface AWSClientInitializationOptions {
+    /**
+     * Custom client branding information.
+     */
+    product?: {
+        name: string
```
awschristou, line 52: Are spaces allowed, or are there any restrictions on the field values?
If so, can these be stated in the comments

runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,41 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
+}
+
+/**
+ * Extended AWS Runtimes InitializeParams interface,
+ * sent from Client to Server as part of [`initialize`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) request
+ */
+export interface InitializeParams extends _InitializeParamsBase {
+    initializationOptions?: {
+        [key: string]: any
+        aws: AWSClientInitializationOptions
```
awschristou, line 75: I'd like to suggest `awsClient`, as this field seems to describe the client. In the future if there was something "AWS" specific that needed to be passed along (like a directive along the lines of "prefer doing this when using service foo"), I'm under the impression this would go in a separate object. But if you intend for all of that to go into `AWSClientInitializationOptions` I have no objection. AWSClientInitializationOptions would simply become a catch-all over time.

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -65,4 +81,31 @@ export class LspRouter {
...
+        // Compute client-specific suffix
+        const { product, platform, clientId } = params || {}
+
+        if (product?.name && product?.version) {
+            items.push(`${product.name}/${product.version}`)
+        }
+
+        if (platform?.name && platform?.version) {
+            items.push(`${platform.name}/${platform.version}`)
+        }
```
awschristou, line 103: If name or version is missing but the other is present, you should still emit the data you do have. I suggest replacing the missing value with some token like "UNKNOWN" or "MISSING"

awschristou: On second thought, you may want to check all fields for null and swap them, so you don't emit something like "USER_AGENT_PREFIX /1.2.3" (in this example this.version is present but this.name is absent). I assume this would trip up any parsers or queries consuming the agent.

awschristou (commented):
This looks good to me. I haven't tried making the client emit this structure, so there's a chance we may encounter some unknown problem there, but I think this is unlikely.

runtimes/runtimes/lsp/router/lspRouter.ts
rli: should we just let the client build this and pass it as-is?

viktorsaws: > should we just let the client build this and pass it as-is?

My idea with building it in runtime is to have a single source of truth for the format based on expected input, so it is predictable.

I considered making clients responsible for passing the suffix, which will make this implementation much simpler.

viktorsaws: I revised proposed protocol, and simplified it to pass only userAgentSuffix from the client. Some data will still be attached in Runtime, but logic is simpler

awschristou: > should we just let the client build this and pass it as-is?

I'm concerned if user agent strings need to be in a specific format. This wont scale well, because each client may/will diverge, and then there will be change requests to fix this in both clients and the back-end as mistakes occur in what is passed along as a pre-built string.

awschristou (commented):
Looks good to me

runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,47 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
+        /**
+         * Product name, may contain spaces.
+         */
+        name: string
+        version: string
+    }
+    /**
+     * Client host platform information.
+     */
+    platform?: {
```
kmile, line 61: What is the purpose of platform? Is the "client host platform" ever different from the platform the runtime runs on?

runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,47 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
 }
 
+/**
+ * Extra Client information passed during initialization handshake from Client to Server.
+ */
+export interface AWSClientInitializationOptions {
+    /**
+     * Custom client branding information.
+     */
+    product?: {
```
kmile, line 51: Is this information a duplicate of the `clientInfo` in the `initializeRequest` itself? Can they be different?

runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,47 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
+        /**
+         * Platform name, may contain spaces.
+         */
+        name: string
+        version: string
+    }
+    /**
+     * Custom Client ID value, set based on client extension internal business logic.
+     */
+    clientId?: string
```
kmile, line 71: What's the difference between `client` and `product`?

runtimes/protocol/lsp.ts
kmile: I'd prefer to rename this `clientInfo` instead of `aws` since that's actually what it captures? I assume we only have this field here because the LSP client in e.g., Visual Studio does not allow us to override the proper `clientInfo` in the `initializeRequest`?

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -42,7 +45,19 @@ export class LspRouter {
...
         }
 
-        let responsesList = await Promise.all(this.servers.map(s => s.initialize(params, token)))
+        let responsesList = await Promise.all(
+            this.servers.map(s =>
+                s.initialize(
+                    {
+                        ...params,
+                        awsRuntimeMetadata: {
+                            customUserAgent: this.getUserAgent(params.initializationOptions?.aws),
```
kmile, line 54: I'm curious why we ask for a lot of client specific identification, something that is useful in many places (for instance https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/codeWhispererServer.ts#L332) and hide it inside a specific User Agent, something Servers might want to compute themselves.

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -65,4 +80,39 @@ export class LspRouter {
             }
         }
     }
+
+    getUserAgent = (opts: AWSClientInitializationOptions = {}): string => {
```
kmile, line 84: Should this be the responsibility of the Servers? Makes this change a lot simpler and smaller in scope.

runtimes/protocol/lsp.ts
viktorsaws: `platform` here refers to the IDE or Editor, in which Client runs. I'll clarify the naming, platform maybe not the best.

From language server process pov, platform is host machine, but not IDE.

runtimes/protocol/lsp.ts
viktorsaws: We can't be sure that standard `clientInfo` values set by IDEs can be customised by Client extensions or if values set by default are enough.
That's why I introduced this `product` field, where Client extensions can set values that are meaningful for them.

runtimes/protocol/lsp.ts
viktorsaws: technically no difference, both point to Client extension integrating with Language Server. `clientId` here is unique identifier set by extension to identify unique installations or sessions of extension. Probable it does not have to be a separate field, and can be nested under `product` object.

runtimes/runtimes/lsp/router/lspRouter.ts
viktorsaws: Servers still receive raw data as part of `InitializationParams`: `params` object here is received from client and extended.

Servers still can compute their own values and choose not to use this precomputed UserAgent, we can't enforce this. The benefit of injecting `customUserAgent` is to own the standard logic, that can be shared by all Servers, avoid duplication of computing the value.

runtimes/runtimes/lsp/router/lspRouter.ts
viktorsaws: I considered this at first. Imo the benefit of owning this logic and passing standard computed value is in producing standard value for all Servers. When each server will compute it's own value, there could be discrepancies and it'll be more difficult to manage all different implementations.

Same scalability/management issue as @awschristou mentioned when this logic is owned by Clients, but reversed to the Servers now https://github.com/aws/language-server-runtimes/pull/193#discussion_r1733180679.

As an alternative, we can export this utility function instead of extending `params` with computed `customUserAgent` value for servers to use, then logic for computing stays in Runtimes, but we don't fiddle with params passed down to servers.

runtimes/protocol/lsp.ts
viktorsaws: I am using `aws` as a namespace to scope all custom extension points we do to standard LSP. I'll nest proposed fields in `aws.clientInfo`, it seems to be logical grouping for these fields based on this thread.

runtimes/protocol/lsp.ts
viktorsaws: I have updated the naming in latest commit https://github.com/aws/language-server-runtimes/pull/193/commits/969bad7d3819d50417800ed3eb0298261066b229. Also splitting platform and client info in 2 fields.

runtimes/protocol/lsp.ts
```diff
@@ -40,6 +41,50 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
...
+ */
+export interface AWSPlatformInitializationOptions {
+    /**
+     * Platform name, may contain spaces.
+     */
+    name: string
+    version: string
+}
+
+export interface AWSInitializationOptions {
```
ege0zcan, line 72: Do I understand correctly that we won't have 'serverInfo' here? so we won't know the name and version of the server

ege0zcan (approved):

runtimes/protocol/lsp.ts
viktorsaws: `serverInfo` is known to Runtime, it is injected with RuntimeProps in build configuration

viktorsaws:
After talking with @kmile, we identified few rough edges in proposed protocol spec. I tried to address them in latest revision. 
We also decided to move userAgent computation out of runtimes to Server implementation, as it is very specific to Servers.

kmile (approved):

PR #201, opened 2024-09-06
Add example of service injection in the runtime

viktorsaws:
## Problem

Some of functionality exposed by runtimes must be served only by single handling service. Current handlers injection model allows to inject a set of `Server` interface implementations, each getting access to number of features (`LSP`, `Telemetry`, `Chat`, `Workspace`, etc.). With current model, each server may handle and respond to incoming requests following their own business logic. 

This poses a problem for cases when a protocol extension can properly function only when served by one handling server and must not be interfered by other servers. One example identified in Auth management feature.

## Solution

One identified solution is add second extension point called `services` to the interface of runtimes functions in `RuntimeProps` and accept implementation handlers for predefined set of protocol handlers. Each service in the list will be unique and will be able to serve it's own predefined set of custom LSP requests. LSP requests are defined in and part of `runtimes/protocol`.

Services implementations can be passed to the runtime at build time in BuildConfiguration file. This way, consumer of resulting language server bundle is controlling with functionality they want to inject into their bundle. The decision where to host business logic will be up to developer.

This PR demonstrates PoC implementation of this mechanism for AuthManagement feature.
* Example of build configuration is in `runtimes/examples/buildConfigurationWithServices.ts`
* Example of business logic handler implementation is in `runtimes/testing/TestAuthManagementService.ts`
* Wiring LSP requests to handers is done in https://github.com/aws/language-server-runtimes/pull/201/files#diff-4bede657896036903a4019c3fd2c3fc4d8d2c1bfb0735ff2d58e4baa67db173dR23-R47
* Final Protocol API is TBD

## License

### DIFF
--- a/runtimes/examples/buildConfigurationWithServices.ts
+++ b/runtimes/examples/buildConfigurationWithServices.ts
@@ -1,0 +1,19 @@
+/**
+ * This is example configuration of creating a bundle with only custom AuthManagementService injected into standalone runtime.
+ * This setup produces a server, that only handles AuthManagement LSP requests.
+ */
+
+import { RuntimeProps } from '@aws/language-server-runtimes/runtimes/runtime'
+import { standalone } from '@aws/language-server-runtimes/runtimes/standalone'
+import { createAuthManagementService } from '../testing/TestAuthManagementService'
+
+const props: RuntimeProps = {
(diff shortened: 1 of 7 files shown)


viktorsaws:
Proposal is discarded after internal team review

PR #202, opened 2024-09-10
Draft. Identity management feature

imykhai:
##  Notes
Draft/first changes for Identity management feature to be implemented by a capability server. 

## License


runtimes/runtimes/webworker.ts
```diff
@@ -160,7 +162,11 @@ export const webworker = (props: RuntimeProps) => {
             },
         }
 
-        return s({ chat, credentialsProvider, lsp, workspace, telemetry, logging, runtime })
+        const identityManagement: IdentityManagement = {
```
dogusata, line 165: could it be a helper function?

dogusata (approved):
Only thing i see is the same `identityManagement` set twice in 2 files. Could it be a helper function which returns the expected `IdentityManagement` ?

runtimes/runtimes/standalone.ts
```diff
@@ -249,7 +251,11 @@ export const standalone = (props: RuntimeProps) => {
                 chat = new BaseChat(lspConnection)
             }
 
-            return s({ chat, credentialsProvider, lsp, workspace, telemetry, logging, runtime })
+            const identityManagement: IdentityManagement = {
+                onListProfiles: handler => lspConnection.onRequest(listProfilesRequestType, handler),
+            }
+
+            return s({ chat, credentialsProvider, lsp, workspace, telemetry, logging, runtime, identityManagement })
```
volodkevych, line 258: The final interface with parameters that we inject becomes confusing for server, in my opinion. Part of the parameters are basic services that servers just consume, like `credentialsProvider`, `workspace`, `telemetry`, logging` and `runtime`. `lsp` and `chat` are specific, because those are LSP-specific and servers need to implement them. `identityManagement` also belongs to the 2nd group. I think we should not continue defining those LSP-specific handlers that server needs to implement here, but rather define them in `extensions` of `lsp`, as we did for `onInlineCompletionWithReferences`.
Keep in mind that with Toolkits common components, this list will grow further.

imykhai:
@dogusata 
> Only thing i see is the same identityManagement set twice in 2 files. Could it be a helper function which returns the expected IdentityManagement ?

they are different. First one is the protocol extension - we define the structure of communication between the client and the bundle/process(/webworker). The second is part of server interface. And as such it will require implementation from the server developer.

runtimes/runtimes/standalone.ts
kmile: I don't think IdentityManagement is really part of `Lsp`, as in, the _Language Server Protocol_. It's definitely a runtime feature and capability that servers can implement, with or without doing anything with programming languages.

kmile: It seems more confusing to me we have a `credentialsProvider` and something that is _effectively_ a credentials provider, but one provides credentials to `Server`s and the other manages credentials for clients.

We used to have an `Auth` feature, is it time to bring that back and merge the two?

runtimes/protocol/identity-management.ts
```diff
@@ -0,0 +1,22 @@
+import { ProtocolRequestType } from './lsp'
+
+export type ListProfilesParams = {}
+export type ListProfilesResult = {
+    profiles?: any[]
+    ssoSessions?: any[]
```
kmile, line 6: Is there now way to scope this down from `any`?

floralph: Below is the original Toolkits design from the TKC: Auth Design doc.  It only takes into account non-legacy SSO profiles/sso-sessions currently, but there is an expectation that it will be extended to support other profile types in the future.  

There are quite a few fields and profile types which would be time-consuming and unnecessary to build out upfront.  I originally envisioned having separate TypeScript types and JSON-RPC params for each profile type to make it easier for consuming code to work with rather than a single profile type with every possible field on it.  The latter would push validation and which fields apply logic into all of the consumers that would likely be implemented inconsistently and potentially drift over time.

The AWS CLI docs probably have the most complete definition of the possible fields.  https://docs.aws.amazon.com/cli/v1/userguide/cli-configure-files.html

```
export type SsoTokenProfileKind = 'SsoToken'

export type ProfileKind = SsoTokenProfileKind

export const ProfileKind = {
    SsoToken: 'SsoToken'
} as const

export interface Section {
    name: string
}

export interface Profile extends Section {
    readonly kind: ProfileKind
    region?: string
}

export interface SsoTokenProfile extends Profile {
    readonly kind: SsoTokenProfileKind
    ssoSessionName: string
}

export interface SsoSession extends Section {
    ssoStartUrl: string
    ssoRegion: string
    ssoRegistrationScopes?: string[]
}

export interface ListProfilesResult {
    profiles?: (Profile | SsoTokenProfile)[]
    ssoSessions?: SsoSession[]
}
```

runtimes/protocol/identity-management.ts
```diff
@@ -0,0 +1,22 @@
...
+}
+export type ListProfilesError = {}
```
floralph, line 8: I'm working through a concept of handling enriched errors now.  I'll submit code/PR to the team in a doc to discuss further.  For now, an empty error type is fine.

runtimes/protocol/identity-management.ts
```diff
@@ -0,0 +1,22 @@
...
+export const listProfilesRequestType = new ProtocolRequestType<
+    ListProfilesParams,
+    ListProfilesResult,
+    never,
+    ListProfilesError,
+    void
+>('aws/identity/listProfiles')
```
floralph, line 15: If we include the profile-related calls (ListProfiles and UpdateProfile, maybe more in the future) in the identity runtime feature rather than a separate profile feature, it will need to throw a not implemented error in browser runtimes.  

While we could theoretically introduce the concept of profiles in browsers and store in a browser-based storage rather than shared config, that is an entirely separate discussion out of scope for this project.

runtimes/runtimes/standalone.ts
floralph: There is an Auth feature.  

The original proposal in the TKC: Auth Design was to extend that Auth feature to respond to the JSON-RPC requests to handle client identity management functions (such as SSO login).  The Auth feature exposes the CredentialsProvider interface injected into the LSP servers which offers only limited functionality that the servers need.  The client functions such as setting and erasing the current creds/token are only exposed to clients via the JSON-RPC interface, not to LSP servers via the CredentialsProvider interface.

The choice was made to move the client functions into an LSP server instead for clear lines of code ownership and to fit the expected workflow for 3rd-party contributors among other things.  I do get volodkevych@'s point about separate responsibilities though as most of these injected features "support the server" (e.g. I need a secret to call this API, what is it?) whereas the identityManagement feature primarily creates a means to expose non-LSP API calls to LSP servers to handle (similar to what the Lsp server-interface does for LSP).

https://github.com/aws/language-server-runtimes/blob/main/runtimes/protocol/auth.ts
https://github.com/aws/language-server-runtimes/blob/main/runtimes/runtimes/auth/auth.ts
https://github.com/aws/language-server-runtimes/blob/main/runtimes/server-interface/auth.ts

Something like:

```
s({ chat, credentialsProvider, lsp, workspace, telemetry, logging, runtime, extensions: { identityManagement: identityManagement } })
```

It makes sense to avoid further bloating the parameter list and for most servers, they can probably safely ignore the extensions arg entirely.  Though "extensions" is ambiguous.  Maybe a different name such as `apis` or `apiExtensions` would be better fit.

imykhai: We've had some offline discussions about this. To summarize:

There is an agreement that the `identityManagement` feauture doesn't belong to the `lsp`. 
Finding the long-term approach on how we separate `apiExtenstions`/`features` (`what servers use` vs `what servers implement`) is better to do separately from this PR because we don't have an agreement on whether we need to change the interface at all and the proposed ideas are backward incompatible anyway. 

Thus I propose to keep the `identityManagement` inside the `features` object in this PR.

runtimes/protocol/identity-management.ts
imykhai: > While we could theoretically introduce the concept of profiles in browsers and store in a browser-based storage rather than shared config, that is an entirely separate discussion out of scope for this project.

that's the idea here: in the runtimes package we define the protocol and server interface, and it's up to the server implementation whether it uses `WebStorage` or `FileSystem` or anything else. With that being said, ideally, server implementations should be runtime-agnostic with actual runtime-specific features injected into them via common interface.

floralph: I did a little exploration in the space of having the runtime package implement the runtime-specific file access regarding shared config.  There were some challenges such as the paths are specific (~/.aws/config & credentials) and having to access environment variables which the browser cannot do either.  Given that was just requirements for shared config and I'm sure other file/host access requirements will differ, I gave up on that approach and went back to implementing file access in the server for now.  The implementation does model native storage vs web storage so they can be swapped out within the server at least.

PR #210, opened 2024-09-17
Create new extension for getAvailableServerConfigurations

ege0zcan:
## Description
Creating a new extension to define a request made from clients to servers to get a list of available configuration values. This will expose a way for clients to discover server-side configurations and their values 

## License


runtimes/protocol/getAvailableServerConfigurations.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const getAvailableServerConfigurationsRequestType = new ProtocolRequestType<
+    GetAvailableServerConfigurationsParams,
+    any[],
+    never,
+    void,
+    void
+>('aws/getAvailableServerConfigurations')
+
+export interface GetAvailableServerConfigurationsParams {
```
rahmaniaam, line 11: Config params already exists https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_configuration. Can we use/extend those instead of creating a custom one ourselves?

runtimes/runtimes/standalone.ts
```diff
@@ -246,6 +247,9 @@ export const standalone = (props: RuntimeProps) => {
                 onHover: handler => lspConnection.onHover(handler),
                 onSignatureHelp: handler => lspConnection.onSignatureHelp(handler),
                 extensions: {
+                    onGetAvailableServerConfigurations(handler) {
```
viktorsaws, line 250: I'd suggest to remove `Server` from `getAvailableServerConfigurations` naming. When I read `ServerConfiguration`, I think of it as something that allows me to configure server, which is not the intent here.

runtimes/protocol/getAvailableServerConfigurations.ts
viktorsaws: I think we should not mix arguments with existing LSP config params, as intent is slightly different. We probably won't need `scopeUri`, and `section` values might not overlap. Also we'll have more flexibility evolving the payload separately from standard LSP, if needed

runtimes/protocol/lsp.ts
```diff
@@ -107,5 +108,14 @@ export interface InitializeResult extends _InitializeResultBase {
...
+        configurationProvider?: ConfigurationOptions
     }
 }
+
+/**
+ * Configuration options for AWS Runtimes
+ * Sent back to the client to signal available configuration values
+ */
+export interface ConfigurationOptions {
+    configurations: string[]
```
viktorsaws, line 120: Is this suppose to be a list of supported `sections`, to which server can reply with configurations? Then i'd suggest to change `configurations` to `sections` to be more explicit

runtimes/protocol/getAvailableServerConfigurations.ts
```diff
@@ -0,0 +1,13 @@
+import { ProtocolRequestType } from './lsp'
+
+export const getConfigurationFromServerRequestType = new ProtocolRequestType<
+    GetConfigurationFromServerParams,
+    any[],
```
pras0131, line 5: Rather than `any[]`, can we create a interface for response? This will make it clear to anyone reading it what the response structure is going to be like.

runtimes/protocol/getConfigurationFromServer.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const getConfigurationFromServerRequestType = new AutoParameterStructuresProtocolRequestType<
+    GetConfigurationFromServerParams,
+    LSPAny,
+    never,
+    void,
+    void
+>('aws/getConfigurationFromServer')
+
+export interface GetConfigurationFromServerParams {
```
viktorsaws, line 11: nit, can be done in follow-up: let's start adding documentation to protocol changes to explain purpose of params like this for servers and clients developers.

types/getConfigurationFromServer.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Configuration options for AWS Runtimes
+ * Sent back to the client to signal available configuration values
+ */
+export interface ConfigurationOptions {
```
viktorsaws, line 5: Does this object really need to be in `types` package and not `protocol`?

runtimes/protocol/getConfigurationFromServer.ts
```diff
@@ -0,0 +1,13 @@
+import { AutoParameterStructuresProtocolRequestType, LSPAny } from './lsp'
+
+export const getConfigurationFromServerRequestType = new AutoParameterStructuresProtocolRequestType<
+    GetConfigurationFromServerParams,
+    LSPAny,
+    never,
+    void,
```
viktorsaws, line 7: Should server be able to return errors?

viktorsaws (approved):

runtimes/protocol/getConfigurationFromServer.ts
ege0zcan: tbh I couldn't think of a reason why servers should return errors in the config scenario, do you have any scenario in mind?

types/getConfigurationFromServer.ts
ege0zcan: I was also conflicted, seeing as `ChatOptions` is also defined in types https://github.com/aws/language-server-runtimes/blob/14c7fc6c782faa7cf22f39d889f659d4faa7c8a6/runtimes/protocol/lsp.ts#L109 
I wanted to keep consistency so I moved it to types

pras0131 (approved):

types/getConfigurationFromServer.ts
viktorsaws: Yeah, I'd keep scope of changes small tbh, and don't move stuff to types unnecessarily. It was originally meant to host types around Chat feature. Maybe keeping LSP `Options` types there is not right and we'll move `ChatOptions` out as well.

PR #214, opened 2024-09-23
Add getAllTextDocuments method to Workspace feature

viktorsaws:
## Problem
Workspace feature does not expose function to get all open files in the IDE.

## Solution

Added new method `getAllTextDocuments` which returns all documents managed by LSP server.

## License

### DIFF
--- a/runtimes/server-interface/workspace.ts
+++ b/runtimes/server-interface/workspace.ts
@@ -16,7 +16,7 @@ interface Dirent {
  */
 export type Workspace = {
     getTextDocument: (uri: string) => Promise<TextDocument | undefined>
-    getAllTextDocuments: () => TextDocument[]
+    getAllTextDocuments: () => Promise<TextDocument[]>
     getWorkspaceFolder: (uri: string) => WorkspaceFolder | null | undefined
     fs: {
         copy: (src: string, dest: string) => Promise<void>
--- a/runtimes/server-interface/workspace.ts
+++ b/runtimes/server-interface/workspace.ts
@@ -16,6 +16,7 @@ interface Dirent {
  */
 export type Workspace = {
     getTextDocument: (uri: string) => Promise<TextDocument | undefined>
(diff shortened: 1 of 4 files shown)


runtimes/server-interface/workspace.ts
```diff
@@ -16,6 +16,7 @@ interface Dirent {
  */
 export type Workspace = {
     getTextDocument: (uri: string) => Promise<TextDocument | undefined>
+    getAllTextDocuments: () => TextDocument[]
```
kmile, line 19: Is there a reason servers can't watch `https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didOpen` and keep track of the interesting ones themselves?

viktorsaws: They of course can, but I think it's not scalable approach. It pushes unnecessary complexity to each Server implementation - if tomorrow server developer decides they need access files, they need to reimplement caching and monitoring. Here, this API is provided from central place, which keeps track of open documents anyways.

viktorsaws: I think I'd rather convert `getAllTextDocuments` to `async`, for future extensibility and to keep it consistent with rest of methods in Workspace feature.

ege0zcan: > I think I'd rather convert `getAllTextDocuments` to `async`, for future extensibility and to keep it consistent with rest of methods in Workspace feature.

I agree with this approach

ege0zcan:
```
returns all documents managed by LSP server.
```
what does this mean in practice? Does it return for example all files in a workspace that are of type that is watched by the server?

viktorsaws:
> ```
> returns all documents managed by LSP server.
> ```
> 
> what does this mean in practice? Does it return for example all files in a workspace that are of type that is watched by the server?

All files registered in this language server instance by client or watched by server, yes.

ege0zcan (approved):

PR #215, opened 2024-09-25
Add routing to initialized and didChangeConfiguration notifications

ege0zcan:
## Description

- Added routing for `didChangeConfiguration` and `initialized` notifications so that all servers in a bundle will get the notifications instead of the last server that registered a handler 

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/lspRouter.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.ts
@@ -1,8 +1,11 @@
 import {
     CancellationToken,
+    DidChangeConfigurationNotification,
+    DidChangeConfigurationParams,
     ExecuteCommandParams,
     GetConfigurationFromServerParams,
     getConfigurationFromServerRequestType,
+    InitializedParams,
     InitializeError,
     InitializeParams,
     InitializeResult,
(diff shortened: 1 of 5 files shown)


runtimes/runtimes/webworker.ts
```diff
@@ -128,18 +126,10 @@ export const webworker = (props: RuntimeProps) => {
...
-            onInitialized: handler =>
-                lspConnection.onInitialized(p => {
-                    const workspaceCapabilities = lspRouter.clientInitializeParams?.capabilities.workspace
-                    if (workspaceCapabilities?.didChangeConfiguration?.dynamicRegistration) {
-                        // Ask the client to notify the server on configuration changes
-                        lspConnection.client.register(DidChangeConfigurationNotification.type, undefined)
-                    }
-                    handler(p)
-                }),
+            onInitialized: lspServer.setInitializedHandler,
```
viktorsaws, line 129: Why allowing Server to set `InitializedHandler` is needed? 
How this will impact server implementations, will every server now need to provide this handler with logic removed above?

viktorsaws: Please ignore, I noticed that this logic is moved to LSPRouter, which runs it before calling handler the old way. So this is just refactoring and does not change current implementation, right?

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -69,6 +74,24 @@ export class LspRouter {
         }
     }
 
+    didChangeConfiguration = (params: DidChangeConfigurationParams): void => {
+        for (const s of this.servers) {
+            s.sendDidChangeConfigurationNotification(params)
```
viktorsaws, line 79: Good catch!

viktorsaws (approved):

runtimes/runtimes/webworker.ts
ege0zcan: Yes, as part of this change I moved the logic from runtimes to the LSP router

PR #220, opened 2024-10-04
Add export and comments about how identity-management should be exported

floralph:
Exports protocol/identity-management through server-interace/identity-management because all of the request/result/error types are needed by the aws-lsp-identity server as it handles all of them (it is not an LSP server, despite the name).  Comments were added to explain the intention here as well as in a separate PR in the language-servers repo.


runtimes/server-interface/identity-management.ts
```diff
@@ -18,6 +18,12 @@ import {
 } from '../protocol/identity-management'
 import { RequestHandler } from '../protocol'
 
+// As the aws-lsp-identity server needs access to all the param/result/error types for requests/notifications because
+// it is solely responsible for handling them, export all of the protocol/identity-management to server-interface/identity-management.
+// Do not export any types here to server-interface/index.ts as only the aws-lsp-identity server should be using these types.
```
volodkevych, line 23: Should it really be the limitation given we inject `identityManagement` [here](https://github.com/aws/language-server-runtimes/blob/main/runtimes/runtimes/standalone.ts#L264) into every server?

floralph: It may be injected into the factories, but no servers are importing it via the "features" parameter, so it's just ignored.  Of course that is because it has just been added, but other servers don't need those requests/notifications, so they should just continue ignoring it.

Unless subsequent identity servers are created, there is no reason for any server outside of aws-lsp-identity to import IdentityManagment as they won't be listening for those messages.  I would thoroughly question why new identity servers are created instead of modifying the existing one.

Ultimately, it's mostly to prevent polluting the top-level protocol and server-interface namespaces.  It's trivial to tack identity-management to the end of import statements.  This is going to be the same situation when more one-off APIs are added in the future.  This goes back to the design review discussion around how to add one-off APIs (i.e. only one server needs them) while mitigating impact to the overall namespaces (i.e. these things shouldn't be global if they're not shared).

viktorsaws: For the protocol defined in this `Runtimes` project we (internally) decided not to treat one-off APIs in any special way, and simply treat them all as additions to the protocol.
So we effectively are separating one-off implementation of some APIs from protocol: the `aws-lsp-identity` is one possible implementation of server for `identityManagement` protocol. Ensuring that this implementation is the only one serving requests defined in `identityManagement` interface is controlled in server artifact build configuration at build time.

I think we shouldn't even mention `aws-lsp-identity` in comments here, since technically other implementations are possible.

volodkevych: I would prefer to have `identityManagement` interface exported as others for consistency. Having `identityManagement` interface exported differently does not prevent other servers from using it. If we need to provide  support for one-off APIs on runtime level, it should be investigated/designed separately. As Viktor said, server artifact build configuration defines servers to be used in bundle, so by manually defining it you can ensure only one `identityManagement` server used.

floralph: Removed comments and added identity-management export back to index.ts

volodkevych (approved):

PR #231, opened 2024-10-15
Notifications protocol

volodkevych:
## Problem

New custom protocol was requested for notifications. Also, the server interface was extended to give possibility implement notification servers.

## Solution

- New notifications protocol was created
- Servers interface was extended to support notifications
- `InitializeParams` were extended with data directory - it will be used to store notification cache file
- README was extended

## License


runtimes/runtimes/standalone.ts
```diff
@@ -183,6 +194,12 @@ export const standalone = (props: RuntimeProps) => {
             },
         }
 
+        const notifications: Notifications = {
+            showNotifications: params => lspConnection.sendNotification(showNotificationsRequestType.method, params),
```
ege0zcan, line 198: Given we are defining, `showNotificationsRequestType`, wouldn't it work to do: `lspConnection.sendNotification(showNotificationsRequestType, params)`

Not sure if it changes things, but I vaguely remember that it could sometimes make a difference

runtimes/protocol/notifications.ts
```diff
@@ -0,0 +1,73 @@
...
+ * sends to the client to provide notifications to show to customers.
+ */
+export const showNotificationsRequestType = new ProtocolNotificationType<NotificationsParams, void>(
+    'aws/window/showNotifications'
+)
+
+/**
+ * notificationFollowupRequestType defines the custom method that the language client
+ * sends to the server to provide asynchronous customer followup to notification shown.
+ * This method is optional per notification, as not notifications require followup.
```
ege0zcan, line 69: nit, typo: `not all notifications require`

ege0zcan (approved):

runtimes/protocol/lsp.ts
```diff
@@ -85,6 +85,10 @@ export interface AWSInitializationOptions {
      * Additional clientInfo to extend or override default data passed by LSP Client.
      */
     clientInfo?: ExtendedClientInfo
+    /**
+     * Client data directory, used to store LSP server-specific data.
```
viktorsaws, line 89: Can you clarify more in documentation the purpose of this field, and what are the expectation for Servers and Clients about using it?

runtimes/protocol/notifications.ts
```diff
@@ -0,0 +1,73 @@
+import { MessageType, ProtocolNotificationType } from './lsp'
+
+export interface NotificationContent {
+    readonly text: string
+    readonly title?: string
+}
+
+export namespace FollowupNotificationActionType {
```
viktorsaws, line 8: Why `FollowupNotificationActionType` should be separate namespace from `NotificationActionType`?

volodkevych: It's expected to be used for `NotificationFollowupParams`.

runtimes/protocol/notifications.ts
volodkevych: done

ege0zcan (approved):

runtimes/runtimes/standalone.ts
floralph: I ran into a problem attempting to use the form like `lspConnection.sendNotification(showNotificationsRequestType, params)` in the vscode sample client.  I had to use the `.method` form.

There are more details in one of my recent PRs:  https://github.com/aws/language-servers/pull/489#discussion_r1795690421

runtimes/README.md
```diff
@@ -223,6 +223,19 @@ The Identity Management feature is designed to centralize the management of auth
 
 Complete Identity Management parameter and result interfaces can be found in [identity-management.ts](src/protocol/identity-management.ts)
 
+### Notification
+
+The notification feature can be used to send custom customer-facing notifications to clients. Notifications can contain actions, like show URL, but also followup actions, like request customer acknowledgement. When customer reacts to followup actions, asyncronous notification can be sent from client to server to notify server about this.
```
floralph, line 228: nit: type asyncronous -> asynchronous

Would it be helpful to compare/contrast with the built-in LSP ShowMessage and/or give rationale for when to use which one?  

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage

runtimes/protocol/lsp.ts
```diff
@@ -85,6 +85,13 @@ export interface AWSInitializationOptions {
      * Additional clientInfo to extend or override default data passed by LSP Client.
      */
     clientInfo?: ExtendedClientInfo
+    /**
+     * Client data directory, used as an application data directory to store LSP server-specific data.
+     * LSP servers are expected to use the directory for storing their configuration files,
+     * temporary data, and user-specific settings.
+     * The directory is extected to maintain data across updates and re-installations.
+     */
+    dataDirectory?: string
```
floralph, line 94: "Data" is a bit nebulous, is there a crisper name for this parameter?  Something like `lspHome` or `serverFolder`?

nit:  In searching the LSP spec, "folder" is the terminology used in the spec rather than "directory".  Consider renaming for consistency.

Is the structure under that folder covered somewhere else?  i.e. What will enforce each LSP server to isolate its data in a sane and consistent manner (not just start dumping generically named files under the folder path provided)?  

What happens if multiple IDEs use the same path (say VSC and JetBrains installed on the same machine and both pass `$HOME/awsLanguageServers` or something like that)?  

Is the expectation that a configuration or data change in one IDE would impact all other tools consuming the language servers on that host?

runtimes/protocol/notification.ts
```diff
@@ -0,0 +1,69 @@
...
+    | typeof FollowupNotificationActionType.Acknowledge
+
+export interface NotificationAction {
+    readonly text: string
+    readonly type: NotificationActionType
+}
+
+export interface UrlAction extends NotificationAction {
+    readonly type: typeof NotificationActionType.Url
+    readonly url: string
```
floralph, line 31: nit:  LSP spec uses "uri" instead of "url" throughout.  Consider changing for consistency.

runtimes/protocol/notification.ts
```diff
@@ -0,0 +1,69 @@
...
+    | typeof NotificationActionType.Url
+    | typeof NotificationActionType.Marketplace
+    | typeof FollowupNotificationActionType.Acknowledge
+
+export interface NotificationAction {
+    readonly text: string
+    readonly type: NotificationActionType
+}
+
+export interface UrlAction extends NotificationAction {
```
floralph, line 29: nit:  Consider naming `NotificationUriAction` to avoid ambiguity when consuming this type in code.

runtimes/protocol/notification.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface MarketplaceAction extends NotificationAction {
+    readonly type: typeof NotificationActionType.Marketplace
+}
+
+export interface AcknowledgeRequestAction extends NotificationAction {
+    readonly type: typeof FollowupNotificationActionType.Acknowledge
+}
+
+export interface NotificationParams {
+    readonly id?: string
```
floralph, line 43: Do we really want to leave ID as optional?  For follow-up purposes and any other reason to perform an action/update against a notification received, we'd need an ID, correct?  If notifications don't have IDs in their source, it's trivial for a server to just generate a UUID or hash and just ignore it if not needed.  The concern is it seems reasonable that we will at some point have functionality that requires and ID and it would be better for servers to implement that as a requirement upfront.

runtimes/protocol/notification.ts
```diff
@@ -0,0 +1,69 @@
...
+    readonly url: string
+}
+
+export interface MarketplaceAction extends NotificationAction {
+    readonly type: typeof NotificationActionType.Marketplace
+}
+
+export interface AcknowledgeRequestAction extends NotificationAction {
+    readonly type: typeof FollowupNotificationActionType.Acknowledge
+}
```
floralph, line 40: How are these to be used/why define them?

volodkevych: Exactly how they are used now - to pass the type of action to the client, so that client can decide what to do with it.

runtimes/protocol/notification.ts
volodkevych: `Id` is currently needed for Toolkit notifications only. If there is followup expected, `id` would be needed. However, followup is not a common case. I would prefer to leave this optional.

runtimes/protocol/notification.ts
volodkevych: We already use both options throughout the repo.

runtimes/README.md
volodkevych: Will do in followup PR.

runtimes/protocol/lsp.ts
volodkevych: Will rename in followup PR. Also, will extend comment with more details. I think the recommendation for the client would be to provide folder that is specific per-client/IDE.

volodkevych: About LSP server isolating its data within the folder - we can think about exposing `getClientDataDirPath` from `fs` feature as an abstraction to facilitate proper usage of the folder.

volodkevych: Also, keep in mind that this folder is optional in parameters. I introduced it for cases like [globalStorageUri](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage) in vscode. If not specified, runtime will define it - with subfolder per server per client.

runtimes/protocol/notification.ts
floralph: Ah, I see now.  They would be passed in NotifactionParams.actions as concrete NotificationAction types.  Didn't catch that on the first pass.

PR #232, opened 2024-10-17
Renamed dataDirectory field to a more descriptive name

volodkevych:
## Problem

`dataDirectory` requires more descripitive and consistent name.

## Solution

- Renamed field in InitializeParams to a more descriptive and consistent name based on [feedback](https://github.com/aws/language-server-runtimes/pull/231#discussion_r1803524400). It's a breaking change, but the field is unused yet.
- Improved comments

## License

### DIFF
--- a/runtimes/README.md
+++ b/runtimes/README.md
@@ -225,9 +225,11 @@ Complete Identity Management parameter and result interfaces can be found in [id
 
 ### Notification
 
-The notification feature can be used to send custom customer-facing notifications to clients. Notifications can contain actions, like show URL, but also followup actions, like request customer acknowledgement. When customer reacts to followup actions, asyncronous notification can be sent from client to server to notify server about this.
+The notification feature can be used to send custom customer-facing notifications to clients. Notifications can contain actions, like show URL, but also followup actions, like request customer acknowledgement. When customer reacts to followup actions, asynchronous notification can be sent from client to server to notify server about this.
 
 Notifications should be used in rare / exceptional cases that require customer attention (like some change happened or action recommended), but are not blocking the main flow. Clients can decide to throttle notifications, if too many are sent.
+
+Consider using LSP [ShowMessage notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage) instead, if your notification does not require actions or followup actions.
 
 #### Feature Specification
 
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -86,12 +86,16 @@ export interface AWSInitializationOptions {
      */
     clientInfo?: ExtendedClientInfo
     /**
-     * Client data directory, used as an application data directory to store LSP server-specific data.
-     * LSP servers are expected to use the directory for storing their configuration files,
-     * temporary data, and user-specific settings.
-     * The directory is extected to maintain data across updates and re-installations.
+     * Client data folder, used as an application data folder to store LSP server-specific data.
+     * LSP servers are expected to use the folder for storing their configuration files,
+     * temporary data, and user-specific settings. Servers should create subfolder
+     * within the client data folder for storing server-specific data.
+     *
+     * The folder is expected to be created by client, separate folder per application
+     * (e.g. multiple IDEs on a client machine should not share the same folder),
+     * and is extected to maintain data across updates and re-installations.
      */
-    dataDirectory?: string
+    clientDataFolder?: string
 }
 
 /**


runtimes/protocol/lsp.ts
```diff
@@ -86,12 +86,16 @@ export interface AWSInitializationOptions {
...
+     * LSP servers are expected to use the folder for storing their configuration files,
+     * temporary data, and user-specific settings. Servers should create subfolder
+     * within the client data folder for storing server-specific data.
+     *
+     * The folder is expected to be created by client, separate folder per application
+     * (e.g. multiple IDEs on a client machine should not share the same folder),
+     * and is extected to maintain data across updates and re-installations.
      */
-    dataDirectory?: string
+    clientDataFolder?: string
```
viktorsaws, line 98: Since this folder is expected to store server-managed data, would it be better to call it `serverDataFolder` instead?

volodkevych: This is still client-specific folder. We'll need server subfolders for servers. I still prefer `clientDataFolder`.

viktorsaws: I'm fine with this naming, approved

viktorsaws (approved):

runtimes/README.md
```diff
@@ -225,10 +225,12 @@ Complete Identity Management parameter and result interfaces can be found in [id
 
 ### Notification
 
-The notification feature can be used to send custom customer-facing notifications to clients. Notifications can contain actions, like show URL, but also followup actions, like request customer acknowledgement. When customer reacts to followup actions, asyncronous notification can be sent from client to server to notify server about this.
+The notification feature can be used to send custom customer-facing notifications to clients. Notifications can contain actions, like show URL, but also followup actions, like request customer acknowledgement. When customer reacts to followup actions, asynchronous notification can be sent from client to server to notify server about this.
 
 Notifications should be used in rare / exceptional cases that require customer attention (like some change happened or action recommended), but are not blocking the main flow. Clients can decide to throttle notifications, if too many are sent.
 
+Consider using LSP [ShowMessage notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage) instead, if your notification does not require actions or followup actions.
```
floralph, line 232: LSP's ShowMessageRequest allows for MessageActionItems which effectively achieve the same thing.  I'm struggling to think of how this differs and questioning whether ShowMessageRequest is the correct approach.  

The one thing I noticed about ShowMessageRequest is that it appears to be for modal use based on the description in the spec and the fact that is a request that returns the selected action in its response.  While JSON-RPC and LSP state that requests don't need to be completed before sending other requests and notifications, there isn't a recommendation about using requests like long-polling or leaving them open for long periods of time.

The notifications for toolkit (and presumably for implementors) are modeless/asynchronous in nature regarding if and when the user selects an action, so ShowMessageRequest would not be suitable for that use.

What toolkits needs is in the ShowMessage space, but blurs the lines between the ShowMessage notification and the ShowMessage request in that it's fire and forget like the ShowMessage notification, but allows for actions to be supplied like the ShowMessage request.

The combination of ShowMessage notification and ExecuteCommand almost achieve what is needed in the Notifications API.  The missing piece is that there isn't a way in the LSP spec to provide commands via ShowMessage notification.  I also don't see a generic way of providing commands from a server to a client in any other request/notifications.  If there was a ShowMessageModeless notification that passed an array of Commands in its params, that would be what is needed for the Notifications API I think.

I wonder if Larry considered this or an approach more aligned with LSP when designing?

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#command

runtimes/protocol/lsp.ts
```diff
@@ -86,12 +86,16 @@ export interface AWSInitializationOptions {
...
-     * LSP servers are expected to use the directory for storing their configuration files,
-     * temporary data, and user-specific settings.
-     * The directory is extected to maintain data across updates and re-installations.
+     * Client data folder, used as an application data folder to store LSP server-specific data.
+     * LSP servers are expected to use the folder for storing their configuration files,
+     * temporary data, and user-specific settings. Servers should create subfolder
+     * within the client data folder for storing server-specific data.
+     *
+     * The folder is expected to be created by client, separate folder per application
+     * (e.g. multiple IDEs on a client machine should not share the same folder),
```
floralph, line 95: I don't know that it needs to be addressed in the comments here, but we should consider that there is going to be different kinds of data in this folder, some long-lived and shared across all running instances of a particular IDE/destination to short-lived (IDE instance session or Flare process lifetime) and specific to IDE instance/session data.  

Presumably the implementing server will know which of those types of data it is working with.  In the case of the short-lived/specific to instance data, do we have a sufficient way for servers to tie the temporal data to an IDE instance (this is the case even if the Flare server is not shared between IDE instances)?  I see InitializeParams has an optional processId property that can be passed by the client.  That may be sufficient if the client provides it.

floralph (approved):

runtimes/README.md
volodkevych: Why did you make conclusion it's for modal use? They are not mentioning this explicitly. Unfortunately, they also do not tell much about long waiting requests or any timeouts - maybe there are none. There are some short mentions of possibility of reordering. I think the best way to figure out if it actually works would be to test in some test clients. However, I'm not sure if different IDEs can have different behavior.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#messageOrdering

viktorsaws: I guess main challenge with using standard `ShowMessage` is to be sure that all our custom clients (Toolkits) are able to intercept that message and wrap it with needed UI. I imagine there might be situations, when IDE Client does not allow that and those destinations will have degraded/not optimal UX for custom notifications handling

viktorsaws: Also another concern is that it maybe not possible to extend standard LSP Request params for more complex follow-up actions, required by other features

runtimes/protocol/lsp.ts
volodkevych: Temp folder can be better place for short-lived data.

PR #234, opened 2024-10-17
Extend FS feature with write operations

natdurlik:
## Problem
With the addition of the notification protocol #231, the notification server implementation may require the following functions: create a directory, write to a file, append to a file.

## Solution
Extended the workspace `fs` feature with a subset of Node.js file system methods. Currently, the workspace feature is implemented only in the standalone runtime.
Manually tested using a language servers' implementation.

## License

### DIFF
--- a/runtimes/server-interface/workspace.ts
+++ b/runtimes/server-interface/workspace.ts
@@ -1,4 +1,6 @@
+import { FileHandle } from 'fs/promises'
 import { TextDocument, WorkspaceFolder } from '../protocol'
+import { MakeDirectoryOptions, WriteFileOptions } from 'fs'
 
 // Minimal version of fs.Dirent
 interface Dirent {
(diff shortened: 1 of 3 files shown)


floralph (approved):

runtimes/server-interface/workspace.ts
```diff
@@ -27,5 +29,16 @@ export type Workspace = {
         readFile: (path: string) => Promise<string>
         isFile: (path: string) => Promise<boolean>
         remove: (dir: string) => Promise<void>
+        writeFile: (
+            file: FileHandle | string | Buffer | URL,
```
volodkevych, line 33: Should we expose `FileHandle` to the server? It mean cross-platform server would need to depend on type from NodeJs-specific type.

runtimes/server-interface/workspace.ts
```diff
@@ -27,5 +29,16 @@ export type Workspace = {
...
+            data: string | NodeJS.ArrayBufferView,
```
volodkevych, line 34: Same here.

runtimes/server-interface/workspace.ts
natdurlik: Good catch! I've removed the types that we might not need for the notification server implementation. We can easily extend it later if needed.

PR #242, opened 2024-10-23
Extend FS with an ability to get server-specific data folder

natdurlik:
## Problem
Language servers will require the ability to store server-specific data persistently. With the addition of filesystem write operations in #234, there is now a need for a method in the workspace interface that can provide a dedicated application data folder for given server.

## Solution
Extended the Workspace feature with a new `clientDataFolder` method to retrieve the server data folder path. This method takes into account: the operating system, the `InitializeParams` and the server name. If a `clientDataFolder` is specified in `InitializeParams`, it returns this path concatenated with the server name.
Otherwise, it determines the appropriate application data directory based on operating system standards. Within this directory, it constructs path based on LSP `clientInfo` or AWS `clientInfo` parameter from `InitializeParams`. Finally, a server-specific part is appended to the path.

This approach ensures that each language server has its own dedicated space for storing data.

## License

### DIFF
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -15,6 +15,7 @@ import {
     ShowDocumentRequest,
     showNotificationRequestType,
     notificationFollowupRequestType,
+    InitializeParams,
 } from '../protocol'
 import { ProposedFeatures, createConnection } from 'vscode-languageserver/node'
 import {
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
                         })
                     }),
                 getFileSize: path => stat(path),
+                getServerDataFolder: serverName => {
+                    const initializeParams = lspRouter.clientInitializeParams
+                    const clientSpecifiedLocation = initializeParams?.initializationOptions?.aws.clientDataFolder
+                    if (clientSpecifiedLocation) {
+                        return path.join(clientSpecifiedLocation, serverName)
+                    }
+
+                    const clientFolderName = getClientNameFromParams(initializeParams)
+                    const standardizedClientFolderName = standardizeFileName(clientFolderName)
+
+                    const appDataFolder = getPlatformAppDataFolder()
+                    return appDataFolder
+                        ? path.join(appDataFolder, standardizedClientFolderName, serverName)
+                        : path.join(os.homedir(), `.${standardizedClientFolderName}`, serverName)
+
+                    function getPlatformAppDataFolder(): string {
+                        const platform = process.platform
+                        switch (platform) {
+                            case 'win32':
+                                return process.env.APPDATA || path.join(os.homedir(), 'AppData', 'Roaming')
+
+                            case 'darwin':
+                                return path.join(os.homedir(), 'Library', 'Application Support')
+
+                            case 'linux':
+                                return process.env.XDG_DATA_HOME || path.join(os.homedir(), '.local', 'share')
+
+                            default:
+                                return ''
+                        }
+                    }
+
+                    function getClientNameFromParams(initializeParams: InitializeParams | undefined): string {
+                        const clientInfo = initializeParams?.clientInfo
+                        const awsClientInfo = initializeParams?.initializationOptions?.aws.clientInfo
+
+                        return [
+                            awsClientInfo?.name || clientInfo?.name || '',
+                            awsClientInfo?.version || clientInfo?.version || '',
+                            awsClientInfo?.extension.name || '',
+                            awsClientInfo?.extension.version || '',
+                        ]
+                            .filter(Boolean)
+                            .join('_')
+                    }
+
+                    function standardizeFileName(clientFolderName: string): string {
+                        return clientFolderName
+                            .toLowerCase()
+                            .replace(/[^a-zA-Z0-9]/g, '_') // Replace non-alphanumeric characters with an underscore
+                            .replace(/_+/g, '_') // Replace multiple underscore characters with a single one
+                            .replace(/^_+|_+$/g, '') // Trim underscore characters
+                            .slice(0, 255) // Reduce the filename to avoid exceeding filesystem limits
+                    }
+                },
                 getTempDirPath: () =>
                     path.join(
                         // https://github.com/aws/aws-toolkit-vscode/issues/240
--- a/runtimes/runtimes/webworker.ts
+++ b/runtimes/runtimes/webworker.ts
@@ -86,6 +86,7 @@ export const webworker = (props: RuntimeProps) => {
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
             getTempDirPath: () => '/tmp',
+            getServerDataFolder: _serverName => '',
             readFile: _path => Promise.resolve(''),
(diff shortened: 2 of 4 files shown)


runtimes/runtimes/webworker.ts
```diff
@@ -86,6 +86,7 @@ export const webworker = (props: RuntimeProps) => {
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
             getTempDirPath: () => '/tmp',
+            getServerDataFolder: _serverName => '',
```
floralph, line 89: nit:  What do you think of the name getServerDataDirPath for consistency with the other function names declared under fs?

runtimes/runtimes/standalone.ts
```diff
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
                         })
                     }),
                 getFileSize: path => stat(path),
+                getServerDataFolder: serverName => {
+                    const initializeParams = lspRouter.clientInitializeParams
+                    const clientSpecifiedLocation = initializeParams?.initializationOptions?.aws.clientDataFolder
```
floralph, line 187: Can the aws component be null/undefined here?  Even if TypeScript says it's required, it's just parsed JSON.  Does that parsing validate against spec to ensure it exists?

runtimes/runtimes/standalone.ts
```diff
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
...
+                            awsClientInfo?.name || clientInfo?.name || '',
+                            awsClientInfo?.version || clientInfo?.version || '',
+                            awsClientInfo?.extension.name || '',
+                            awsClientInfo?.extension.version || '',
+                        ]
+                            .filter(Boolean)
+                            .join('_')
+                    }
+
+                    function standardizeFileName(clientFolderName: string): string {
```
floralph, line 231: nit: Maybe call this `standardizeDirName` given that if used on a file, it would convert the dot extension separator to an underscore.  While it's safe to use in this context right now, this function has the potential to be lifted to a global library and you wouldn't want to use it as-is against filenames.

runtimes/runtimes/standalone.ts
```diff
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
...
+                                return process.env.APPDATA || path.join(os.homedir(), 'AppData', 'Roaming')
+
+                            case 'darwin':
+                                return path.join(os.homedir(), 'Library', 'Application Support')
+
+                            case 'linux':
+                                return process.env.XDG_DATA_HOME || path.join(os.homedir(), '.local', 'share')
+
+                            default:
+                                return ''
```
floralph, line 213: Should the default here just be os.homedir()?  Seems safer than calling code, if not checking the result, blindly adding a relative path to an empty string and attempting to use it.

runtimes/runtimes/standalone.ts
```diff
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
...
+                        }
+                    }
+
+                    function getClientNameFromParams(initializeParams: InitializeParams | undefined): string {
+                        const clientInfo = initializeParams?.clientInfo
+                        const awsClientInfo = initializeParams?.initializationOptions?.aws.clientInfo
+
+                        return [
+                            awsClientInfo?.name || clientInfo?.name || '',
+                            awsClientInfo?.version || clientInfo?.version || '',
```
floralph, line 223: Is there any expectation that this folder would be used to persist data across updates to the user's IDE and/or extension?  If this isn't expected to hold long-lived persisted values, should it be moved to a temp folder instead?

If this should be version specific:

1. How will data that should be persisted be migrated?
2. What will clean up the long tail of unused folders after updates?

runtimes/runtimes/webworker.ts
natdurlik: I used the "folder" word instead of "dir" to align with the LSP spec terminology. I plan to rename `getTempDirPath` to `getTempFolderPath` in the future changes for aligning our FS interface with Node.js FS interface. Does `getServerDataFolderPath` look good to you?

runtimes/runtimes/standalone.ts
natdurlik: Agreed, I'll update it.

runtimes/runtimes/standalone.ts
natdurlik: Yes I was thinking about that. The default case is different than the rest because we want to append a hidden directory to it. But I agree it is more coherent to return `os.homedir()` and decide how we append the folder name in the caller function, I'll adjust it.

runtimes/runtimes/standalone.ts
volodkevych: I agree, data should be persisted across updates. Also, are you sure we'll need to append both client and extension names? I think we'll end up with long path which risks to hit the limits.

runtimes/runtimes/standalone.ts
```diff
@@ -181,6 +182,61 @@ export const standalone = (props: RuntimeProps) => {
                         })
                     }),
                 getFileSize: path => stat(path),
+                getServerDataFolder: serverName => {
```
volodkevych, line 185: It would be good to unit test this logic.

runtimes/runtimes/standalone.ts
natdurlik: * Yes, you're right - I hadn't considered version updates. 
* We append both the client and extension names because many extensions could use our protocol, which might lead to collisions.
* I've reduced the folder name's max size a bit in the  `standardizeFolderName` function.

runtimes/runtimes/standalone.ts
natdurlik: Refactored this function to a separate file and added unit tests.

runtimes/runtimes/util/serverDataFolder.test.ts
```diff
@@ -0,0 +1,190 @@
...
+                },
+            },
+        }
+        osHomeDirStub = sinon.stub(os, 'homedir').returns('stubHomeDir')
+    })
+
+    afterEach(() => {
+        sinon.restore()
+        process.env.APPDATA = ''
+        process.env.XDG_DATA_HOME = ''
```
volodkevych, line 43: nit: we could improve this by using something like `mock-env` for mock env vars per test

runtimes/runtimes/standalone.ts
volodkevych: "We append both the client and extension names because many extensions could use our protocol, which might lead to collisions." - Agreed, but I think in the latest implementation version you are using client or extension name now. Shouldn't it contain both?

volodkevych (approved):

runtimes/runtimes/webworker.ts
volodkevych: This interface is not part of protocol, so we could go with `dir` here for consistency.

runtimes/runtimes/standalone.ts
natdurlik: It does contain both, if the values are present.

floralph (approved):

PR #247, opened 2024-10-28
Align the FS interface with Node.js FS

natdurlik:
## Problem
Currently, in the standalone Workspace file system, there are some unreasonable default options passed to the Node.js fs functions, which can be confusing to the developers and produce unexpected results.

## Solution
Changed the interface to align with the Node.js interface or documented it explicitly.
* The `remove` method is renamed to `rm`, and default options now align with the default Node.js behavior. Users will need to specify `force` and `recursive` options themselves.
* Both `readFile` and `readdir` methods pass options that aren't default, but their execution should be as expected by users, as these options change the return type. Added comments to additionally mark this.
* The `copy` method renamed to `copyFile` and `options.ensureDir` param to better describe its behavior along with docs comments.
* The `exists` method is changed to use the promise-based API. Now, promise wrapping is unnecessary and it works in the same way.

These are backward-incompatible changes. I'll be adding necessary changes to the language servers to align with this new interface.

## License

### DIFF
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -167,7 +167,7 @@ export const standalone = (props: RuntimeProps) => {
             },
             fs: {
                 copyFile: (src, dest, options?) => {
-                    if (options?.ensureDir !== false) {
+                    if (options?.ensureDir === true) {
                         const destDir = path.dirname(dest)
                         if (!existsSync(destDir)) {
                             mkdirSync(destDir, { recursive: true })
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -50,8 +50,8 @@ import { RuntimeProps } from './runtime'
 
 import { observe } from './lsp'
 
-import { access, mkdirSync, existsSync } from 'fs'
-import { readdir, readFile, rm, stat, copyFile, writeFile, appendFile, mkdir } from 'fs/promises'
+import { mkdirSync, existsSync } from 'fs'
+import { access, readdir, readFile, rm, stat, copyFile, writeFile, appendFile, mkdir } from 'fs/promises'
 import * as os from 'os'
 import * as path from 'path'
 import { LspRouter } from './lsp/router/lspRouter'
@@ -166,7 +166,7 @@ export const standalone = (props: RuntimeProps) => {
                 }
             },
             fs: {
-                copy: (src, dest) => {
+                copyFileWithCreateDir: (src, dest) => {
                     const destDir = path.dirname(dest)
                     if (!existsSync(destDir)) {
                         mkdirSync(destDir, { recursive: true })
@@ -174,12 +174,9 @@ export const standalone = (props: RuntimeProps) => {
                     return copyFile(src, dest)
                 },
                 exists: path =>
-                    new Promise(resolve => {
-                        access(path, err => {
-                            if (!err) resolve(true)
-                            resolve(false)
-                        })
-                    }),
+                    access(path)
+                        .then(() => true)
+                        .catch(() => false),
                 getFileSize: path => stat(path),
                 getServerDataDirPath: serverName => getServerDataDirPath(serverName, lspRouter.clientInitializeParams),
                 getTempDirPath: () =>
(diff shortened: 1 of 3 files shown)


runtimes/server-interface/workspace.ts
```diff
@@ -19,15 +19,25 @@ export type Workspace = {
...
+         * Reads the contents of a directory.
+         * @param {string} path - The path to the directory.
+         * @returns A promise that resolves to an array of Dirent objects.
+         */
         readdir: (path: string) => Promise<Dirent[]>
+        /**
+         * Reads the entire contents of a file.
+         * @param {string} path - The path to the file.
+         * @returns A promise that resolves to the contents of the file as a string.
+         */
```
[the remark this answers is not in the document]
volodkevych: Also, comment should mention that we set default encoding.

runtimes/runtimes/standalone.ts
```diff
@@ -166,20 +166,17 @@ export const standalone = (props: RuntimeProps) => {
                 }
             },
             fs: {
-                copy: (src, dest) => {
+                copyFileWithCreateDir: (src, dest) => {
```
volodkevych, line 169: That's a bit weird name. Should be leave `copyFile` as is in NodeJs and let servers create the folder? Or expose `ensureDir` as an option?

runtimes/server-interface/workspace.ts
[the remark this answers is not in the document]
ege0zcan: `Would it make sense to expose options for this and set 'utf-8' as default` 

Sounds good to me

runtimes/runtimes/standalone.ts
ege0zcan: I think we have the freedom to stay mostly consistent with nodejs while providing some nice features as a built in. This might be a radical suggestion but I think we can keep this function as is 

I feel like exposing ensureDir at this point is redundant

volodkevych: Having `ensureDir` in options is more flexible to support future possible cases. Also, it would give possibility to leave the previous name, which is way better.

runtimes/server-interface/workspace.ts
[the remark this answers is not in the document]
natdurlik: I added the `options` param with `encoding` set to 'utf-8' in the standalone implementation. I don't think it is possible to add a default parameter in the interface. Should we mention it in the interface documentation in that case?

If we were to add `options` as a default parameter later, would we need to change all the usages?

runtimes/runtimes/standalone.ts
natdurlik: I'm inclined towards adding `ensureDir` option, because it describes what function does internally.

volodkevych (approved):

runtimes/runtimes/standalone.ts
```diff
@@ -166,20 +166,19 @@ export const standalone = (props: RuntimeProps) => {
                 }
             },
             fs: {
-                copy: (src, dest) => {
-                    const destDir = path.dirname(dest)
-                    if (!existsSync(destDir)) {
-                        mkdirSync(destDir, { recursive: true })
+                copyFile: (src, dest, options?) => {
```
ege0zcan, line 169: tbh, this also means we are not being consistent with node fs api. There is no options object in the copyFile function and now we are exposing an options object which implicitly has an `ensureDir` field. 

Imo it is better to either keep the copy function (which doesn't have a 1:1 mapping to node and allows us to be more flexible in the behaviour or the function) Or alternatively we expose copyFile as is and expect servers to do the exists, mkdir logic all on the server side. To me the second option looks worse at this point but might be the better option for the longer term

natdurlik: I decided to go with `copyFile` name with `ensureDir` optional parameter set to false. In that case, default behavior of `copyFile` is aligned with Node.js, we provide additional useful feature for creating directory, and can be extended in the future to add optional `mode` parameter like in Node.js. Also, the behavior can be easily deduced from the method's name.

runtimes/server-interface/workspace.ts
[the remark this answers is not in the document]
floralph: Definitely document the default encoding and in comments make it clear that there is no way to get a Buffer back (yes, the return type declaration shows that, but really call it out in comments/docs). 

The NodeJS fs/promises readFile returns a Buffer by default and can support a Buffer if null is passed in for the encoding.  The implementation of Flare's readFile in standalone doesn't allow for Buffers (only returns strings) and the default encoding has changed.  This will be unexpected by implementors familiar with NodeJS.

I think this falls under the AppSec advice to document any behavioral changes.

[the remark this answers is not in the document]
natdurlik: Currently the default value for encoding is mentioned in the method's description. The return type is also specified in the description, additionally to function declaration. In my opinion it should be clear, do you think we need to emphasize it even more?

PR #250, opened 2024-11-06
Notification interface supports dynamic registration

volodkevych:
## Problem

Notification protocol should be an optional capability for clients - it should support dynamic registration via client capabilities.

## Solution

This change introduces `showNotification as a client capability. Client has to explicitly opt in to receive notifications.

In addition, the following changes were introduced:
- notification followup has to be sent per action (array was replaced with single action)
- server interface supports setting server name
- server name is checked for unique value in runtime

## License

### DIFF
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -97,6 +97,14 @@ export interface AWSInitializationOptions {
      * and is extected to maintain data across updates and re-installations.
      */
     clientDataFolder?: string
+    /**
+     * The client signals AWS capabilities it supports.
+     */
+    awsClientCapabilities?: {
+        window?: {
+            showNotification?: boolean
+        }
+    }
 }
 
 /**
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -102,7 +102,7 @@ export interface AWSInitializationOptions {
      */
     awsClientCapabilities?: {
         window?: {
-            showNotification?: boolean
+            notifications?: boolean
         }
     }
 }
--- a/runtimes/runtimes/lsp/router/lspRouter.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.ts
@@ -36,7 +36,7 @@ export class LspRouter {
 
     public get clientSupportsShowNotification() {
         return (
-            this.clientInitializeParams?.initializationOptions?.aws.awsClientCapabilities?.window?.showNotification ??
+            this.clientInitializeParams?.initializationOptions?.aws.awsClientCapabilities?.window?.notifications ??
             false
         )
     }
@@ -49,9 +49,9 @@ export class LspRouter {
 
         let responsesList = await Promise.all(this.servers.map(s => s.initialize(params, token)))
         responsesList = responsesList.filter(r => r != undefined)
-        const errors = responsesList.find(el => el instanceof ResponseError)
-        if (errors) {
-            return errors as ResponseError<InitializeError>
+        const responseError = responsesList.find(el => el instanceof ResponseError)
+        if (responseError) {
+            return responseError as ResponseError<InitializeError>
         }
         const dupServerNames = findDuplicates(responsesList.map(r => (r as PartialInitializeResult).serverInfo?.name))
         if (dupServerNames) {
--- a/runtimes/runtimes/lsp/router/lspRouter.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.ts
@@ -32,6 +32,13 @@ export class LspRouter {
         lspConnection.onRequest(getConfigurationFromServerRequestType, this.handleGetConfigurationFromServer)
     }
 
+    public get clientSupportsShowNotification() {
+        return (
+            this.clientInitializeParams?.initializationOptions?.aws.awsClientCapabilities?.window?.showNotification ??
+            false
+        )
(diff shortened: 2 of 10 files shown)


runtimes/protocol/lsp.ts
```diff
@@ -97,6 +97,14 @@ export interface AWSInitializationOptions {
      * and is extected to maintain data across updates and re-installations.
      */
     clientDataFolder?: string
+    /**
+     * The client signals AWS capabilities it supports.
+     */
+    awsClientCapabilities?: {
+        window?: {
+            showNotification?: boolean
```
viktorsaws, line 105: nit: should this be called just `notifications`, given this capability supports more than showing them, but also follow-ups (and potentially more in future)

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -32,11 +34,34 @@ export class LspRouter {
...
+
     initialize = async (
         params: InitializeParams,
         token: CancellationToken
     ): Promise<InitializeResult | ResponseError<InitializeError>> => {
         this.clientInitializeParams = params
+
+        let responsesList = await Promise.all(this.servers.map(s => s.initialize(params, token)))
+        responsesList = responsesList.filter(r => r != undefined)
+        const errors = responsesList.find(el => el instanceof ResponseError)
```
viktorsaws, line 52: supernit: `.find` returns single element, singular `error` as name is more correct

runtimes/runtimes/lsp/router/util.ts
```diff
@@ -51,3 +51,13 @@ export function mergeObjects(obj1: any, obj2: any) {
     }
     return merged
 }
+
+export function findDuplicates<T>(array: T[]): T[] | undefined {
+    const seen = new Set<T>()
+    const dups = array
+        .filter(a => a !== undefined)
```
viktorsaws, line 58: Non `undefined` values (empty strings, any other non-string falsy values) can pass this filter and result in error in calling code.

runtimes/runtimes/lsp/router/util.ts
```diff
@@ -51,3 +51,13 @@ export function mergeObjects(obj1: any, obj2: any) {
...
+        .filter(function (a) {
+            return seen.size === seen.add(a).size
```
viktorsaws, line 60: nit: this could be simplified with using `Set.has` method for checking seen element. I had to reread this few times to understand 😅

runtimes/server-interface/lsp.ts
```diff
@@ -58,6 +58,17 @@ export type PartialServerCapabilities<T = any> = Pick<
     | 'signatureHelpProvider'
 >
 export type PartialInitializeResult<T = any> = {
+    /**
+     * Information about the server respresented by @type {Server}.
+     * serverInfo is used to differentiate servers internally in the system and is not exposed to the client.
+     */
+    serverInfo?: {
```
viktorsaws, line 65: Shall we use different name for this data? Since this `serverInfo` is different from regular LSP, it's purpose may be a bit confusing. Maybe some alternative like `serverSystemInfo` or `internalServerInfo` would be better?

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -32,11 +34,34 @@ export class LspRouter {
...
+        if (errors) {
+            return errors as ResponseError<InitializeError>
+        }
+        const dupServerNames = findDuplicates(responsesList.map(r => (r as PartialInitializeResult).serverInfo?.name))
```
viktorsaws, line 56: `findDuplicates` perform simple direct value comparison, is this final implementation or will it change?

runtimes/protocol/lsp.ts
volodkevych: This setting will impact only `showNotification` - followups are not possible if there was no notification sent. I got a perception that capabilities in LSP are called after the methods that they enable. But I can rename it for cases when some other methods are added as part of notification.

runtimes/runtimes/lsp/router/util.ts
volodkevych: Although this function is used for strings now, it's not for strings specifically. Empty strings will also be checked for duplicates, which is fine I think.

runtimes/runtimes/lsp/router/lspRouter.ts
volodkevych: We can look to extend it if dedup of objects is ever needed. Now it's working for primitives and objects by ref.

runtimes/server-interface/lsp.ts
volodkevych: I can rename, but I do not like any of those options. LSP server is kind of a composite of feature servers, that's why I liked the idea of borrowing the property name from LSP server level. Also, given feature servers are not exposed to the client directly, their server info is also not exposed.

runtimes/runtimes/lsp/router/util.ts
volodkevych: it's shorter :), but I updated to make it more explicit

runtimes/protocol/lsp.ts
viktorsaws: LSP is not consistent with capabilities naming, unfortunately

viktorsaws (approved):

PR #251, opened 2024-11-07
Add unit tests setup for standalone and webworker runtimes

natdurlik:
## Problem
Currently we need to test changes in the standalone and webworker functions manually using implementations of language servers.
## Solution
Added unit tests setup for both standalone and webworker. Allows to test features passed to servers as part of the initialization.

## License


runtimes/runtimes/util/testingUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+                return target[property as keyof typeof target]
+            }
+            const method = sinon.stub()
+            ;(target as any)[property] = method
+            return method
+        },
+    })
+}
+
+export type Features = {
```
volodkevych, line 28: Let's define it as part of server interface? We keep creating this type/interface in every new place.

runtimes/runtimes/standalone.test.ts
```diff
@@ -0,0 +1,100 @@
...
+                })
+            })
+        })
+
+        describe('Telemetry', () => {
+            describe('emitMetric', () => {
+                it('should use LSP connection telemetry log event', () => {
+                    let metric: any
+                    features.telemetry.emitMetric(metric)
+                    sinon.assert.calledOnceWithExactly(stubConnection.telemetry.logEvent as sinon.SinonStub, metric)
```
volodkevych, line 70: I do not see big value in tests that check function calling another function, to be honest. What kind of tests do you plan to implement with this setup in the future?

runtimes/runtimes/util/testingUtils.ts
natdurlik: Moved `Features` to `/server-interface/server.ts`.

runtimes/runtimes/standalone.test.ts
natdurlik: I just put it as an example. If there is some logic on the runtime side, then tests can be added like that.

volodkevych (approved):

runtimes/runtimes/util/testingUtils.ts
```diff
@@ -0,0 +1,15 @@
+import sinon from 'sinon'
+
+export function createStubFromInterface<T>(): sinon.SinonStubbedInstance<T> & T {
```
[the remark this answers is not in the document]
natdurlik: You are right, there is `stubInterface` in `ts-sinon` that could be used instead.

PR #253, opened 2024-11-13
Add GetSsoTokenResult.updateCredentialsParams to avoid decrypt/recrtypt in destinations

floralph:
For the purposes of calling the Auth.update method in Flare with the results of GetSsoToken, destinations would need to decrypt the accessToken returned and reformat it for Auth.update.  While the ssoToken field is intended for extension and use of a token directly in a destination, there is value in avoiding decrypt/recrypt for destinations that need to call Auth.update.

This PR adds GetSsoTokenResult.updateCredentialsParams to provide the params a destination may provide to the Auth.update API without requiring any tranformation or work from the destination.  This is similar to the LSP ExecuteCommand pattern allowing servers to provide the appropriate arguments to a command so a client can just return what it received without need for additional detail or work.

### DIFF
--- a/runtimes/protocol/identity-management.ts
+++ b/runtimes/protocol/identity-management.ts
@@ -1,4 +1,10 @@
-import { LSPErrorCodes, ProtocolNotificationType, ProtocolRequestType, ResponseError } from './lsp'
+import {
+    LSPErrorCodes,
+    ProtocolNotificationType,
+    ProtocolRequestType,
+    ResponseError,
+    UpdateCredentialsParams,
+} from './lsp'
 
 // Errors
 export const AwsErrorCodes = {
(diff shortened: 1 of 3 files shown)


runtimes/protocol/identity-management.ts
```diff
@@ -169,6 +175,7 @@ export interface SsoToken {
 
 export interface GetSsoTokenResult {
     ssoToken: SsoToken
```
imykhai, line 177: Is `updateCredentialsParams` the only data that are actually going to be used and sent through the LSP client back to the Q Server? 
If so why do we need to keep the original `ssoToken`?

imykhai (approved):

runtimes/protocol/identity-management.ts
floralph: ssoToken should remain for the original purpose of the destination using the bearer token for local calls (from the destination directly to AWS services).  Today that is CodeCatalyst, but other AWS services are expected to support bearer tokens in the future.  The updateCredentialsParams was just added to make it so destinations didn't have to perform work (beyond just making the call) to execute an Auth update call.

PR #259, opened 2024-11-19
Routing for notifications

volodkevych:
## Problem

Notification followup events from client need to be routed to the server of origin.

## Solution

- Routing of notification followup events based on `serverName` + `id`, `id` fields is restored for followup events
- Encoding for web runtime supporting Unicode (should not be directly needed for routing functionality, but I wanted it supported from the very beginning for consistency with NodeJs version)
- Some refactoring in router by adding helper methods to make routing strategy used more explicit
- Unit tests
- Testing in test vscode client - result can be seen here: https://github.com/aws/language-servers/pull/622

## License

### DIFF
--- a/runtimes/runtimes/encoding.ts
+++ b/runtimes/runtimes/encoding.ts
@@ -1,0 +1,34 @@
+export interface Encoding {
+    decode(value: string): string
+    encode(value: string): string
+}
+
+export class WebBase64Encoding implements Encoding {
+    constructor(private window: WindowOrWorkerGlobalScope) {}
+
+    decode(value: string): string {
+        if (!value) {
+            return value
+        }
+        const decoded = this.window.atob(value)
+        // to support Unicode chars
+        return decodeURIComponent(
+            Array.from(decoded)
+                .map(char => {
+                    return '%' + ('00' + char.charCodeAt(0).toString(16)).slice(-2)
+                })
+                .join('')
+        )
+    }
+
+    encode(value: string): string {
+        if (!value) {
+            return value
+        }
+        // to support Unicode chars
+        const converted = encodeURIComponent(value).replace(/%([0-9A-F]{2})/g, (_, arg) => {
+            return String.fromCharCode(parseInt(arg, 16))
+        })
+        return this.window.btoa(converted)
+    }
+}
--- a/runtimes/runtimes/encoding.ts
+++ b/runtimes/runtimes/encoding.ts
@@ -3,30 +3,26 @@ export interface Encoding {
     encode(value: string): string
 }
 
+const HEX_PAD: string = '00'
+const HEX_REGEX: RegExp = /%([0-9A-F]{2})/g
 export class WebBase64Encoding implements Encoding {
     constructor(private window: WindowOrWorkerGlobalScope) {}
 
     decode(value: string): string {
-        if (!value) {
-            return value
-        }
         const decoded = this.window.atob(value)
         // to support Unicode chars
         return decodeURIComponent(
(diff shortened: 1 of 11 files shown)


runtimes/runtimes/encoding.ts
```diff
@@ -0,0 +1,34 @@
...
+                .join('')
+        )
+    }
+
+    encode(value: string): string {
+        if (!value) {
+            return value
+        }
+        // to support Unicode chars
+        const converted = encodeURIComponent(value).replace(/%([0-9A-F]{2})/g, (_, arg) => {
```
Jurredr, line 29: Small nitpick; perhaps nice to add constants for unicode strings just for code clarity purposes? i.e.

```typescript
private readonly HEX_PAD = '00'
private readonly HEX_REGEX = /%([0-9A-F]{2})/g
```

Jurredr (approved):

runtimes/runtimes/encoding.ts
```diff
@@ -0,0 +1,34 @@
+export interface Encoding {
+    decode(value: string): string
+    encode(value: string): string
+}
+
+export class WebBase64Encoding implements Encoding {
+    constructor(private window: WindowOrWorkerGlobalScope) {}
+
+    decode(value: string): string {
+        if (!value) {
```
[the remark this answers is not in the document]
volodkevych: It was to shortcut the case like empty string. But the code works even without that check, so I'll remove it.

runtimes/runtimes/encoding.ts
volodkevych: done

runtimes/runtimes/lsp/router/routerByServerName.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    processFollowup(followupHandler: NotificationHandler<F>, params: F) {
+        if (!params.source.id) {
+            return
+        }
+
+        const sourceId = this.encoding.decode(params.source.id)
+        const id = this.parseServerName(sourceId)
+        if (id?.serverName === this.serverName) {
```
natdurlik, line 39: What if multiple servers have the same server name?

volodkevych: There is check for this in `LspRouter`.

runtimes/runtimes/lsp/router/routerByServerName.ts
```diff
@@ -0,0 +1,57 @@
+import { EventIdentifier, FollowupIdentifier, NotificationHandler } from '../../../protocol'
+import { Encoding } from '../../encoding'
+
+type NotificationId = {
```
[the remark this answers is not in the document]
volodkevych: We set it into `id` field, so I named it accordingly. This is internal to `RouterByServerName` class, so I do not think naming here is a big issue.

runtimes/runtimes/lsp/router/routerByServerName.ts
```diff
@@ -0,0 +1,57 @@
...
+        const sourceId = this.encoding.decode(params.source.id)
+        const id = this.parseServerName(sourceId)
+        if (id?.serverName === this.serverName) {
+            params = {
+                ...params,
+                source: {
+                    id: id.id,
+                },
+            }
+            followupHandler(params)
```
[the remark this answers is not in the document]
volodkevych: No, `LspServer` stores internal server state, like initialization result from a server, which can be used to decide on the routing.

natdurlik (approved):

PR #260, opened 2024-11-21
Support more log levels in runtime logging feature

pras0131:
The PR includes the changes to support more logging levels as existing scenario has only method - which is log that outputs to lspConnection.console.info. The changes add support for error, warn, info, log, debug in the order of their severity. Log level can be decided by the client and when a runtime is executed, it requests Log level from client through LSP getConfiguration. 
Test cases for the changes have been added.

## License

### DIFF
--- a/runtimes/README.md
+++ b/runtimes/README.md
@@ -136,11 +136,16 @@ Design TBD
 
 #### Initialization
 
-TBD
+The runtimes by default supports the logging feature, allowing servers to output logs varied by different log levels. 
 
 #### Feature Specification
 
-TBD
+The log level to be set in the runtime for logging can be decided by the client as part of initialize request, which is the first
+request initiated by client to the LSP server. The log level can be updated dynamically even after server start by triggering `workspace/didChangeConfiguration` 
+notification from the client - which prompts the runtime to fetch the new log level to be set from the client through LSP `getConfigurations` request. 
+1. Params in the initialize request: `InitializeParams.initializationOptions.logLevel`
+2. Configuration request params for requesting log level from runtimes to client: ``{section: "aws.logLevel"}`` 
+
 
 ### Chat
 
--- a/runtimes/runtimes/lsp/router/loggingServer.ts
+++ b/runtimes/runtimes/lsp/router/loggingServer.ts
@@ -2,17 +2,17 @@ import { InitializeParams, InitializeResult, Logging } from '../../../server-int
 import { LspServer } from './lspServer'
 import { Connection } from 'vscode-languageserver/node'
 import { Encoding } from '../../encoding'
-import { DEFAULT_LOG_LEVEL, isValidLogLevel, LoggingImplementation, LogLevel } from '../../util/loggingUtil'
+import { DEFAULT_LOG_LEVEL, isValidLogLevel, DefaultLogger, LogLevel } from '../../util/loggingUtil'
 
 export class LoggingServer {
-    private logger: Logging
+    private logger: DefaultLogger
     private lspServer: LspServer
 
     constructor(
         private lspConnection: Connection,
         private encoding: Encoding
     ) {
-        this.logger = new LoggingImplementation(DEFAULT_LOG_LEVEL as LogLevel, this.lspConnection)
+        this.logger = new DefaultLogger(DEFAULT_LOG_LEVEL as LogLevel, this.lspConnection)
         this.lspServer = new LspServer(this.lspConnection, this.encoding, this.logger)
         this.lspServer.setInitializeHandler(async (params: InitializeParams): Promise<InitializeResult> => {
             this.updateLoggingLevel(params.initializationOptions?.logLevel ?? ('log' as LogLevel))
--- a/runtimes/runtimes/lsp/router/loggingServer.ts
+++ b/runtimes/runtimes/lsp/router/loggingServer.ts
@@ -32,6 +32,7 @@ export class LoggingServer {
 
     private updateLoggingLevel(logLevel: LogLevel) {
         this.logger.level = logLevel
+        this.logger.info(`Logging level changed to ${logLevel}`)
     }
 
     public getLoggingObject(): Logging {
--- a/runtimes/runtimes/lsp/router/loggingServer.ts
+++ b/runtimes/runtimes/lsp/router/loggingServer.ts
@@ -1,0 +1,44 @@
+import { InitializeParams, InitializeResult, Logging } from '../../../server-interface'
+import { LspServer } from './lspServer'
+import { Connection } from 'vscode-languageserver/node'
+import { Encoding } from '../../encoding'
+import { DEFAULT_LOG_LEVEL, isValidLogLevel, LoggingImplementation, LogLevel } from '../../util/loggingUtil'
+
+export class LoggingServer {
+    private logger: Logging
+    private lspServer: LspServer
+
+    constructor(
+        private lspConnection: Connection,
+        private encoding: Encoding
+    ) {
+        this.logger = new LoggingImplementation(DEFAULT_LOG_LEVEL as LogLevel, this.lspConnection)
+        this.lspServer = new LspServer(this.lspConnection, this.encoding, this.logger)
+        this.lspServer.setInitializeHandler(async (params: InitializeParams): Promise<InitializeResult> => {
+            this.updateLoggingLevel(params.initializationOptions?.logLevel ?? ('log' as LogLevel))
+            return {
+                capabilities: {},
+            }
+        })
+        this.lspServer.setDidChangeConfigurationHandler(async params => {
+            const logLevelConfig = await lspConnection.workspace.getConfiguration({
+                section: 'aws.logLevel',
+            })
+            if (isValidLogLevel(logLevelConfig)) {
+                this.updateLoggingLevel(logLevelConfig as LogLevel)
+            }
+        })
+    }
+
+    private updateLoggingLevel(logLevel: LogLevel) {
+        this.logger.level = logLevel
+    }
+
+    public getLoggingObject(): Logging {
+        return this.logger
+    }
+
+    public getLspServer(): LspServer {
+        return this.lspServer
+    }
+}
--- a/runtimes/runtimes/standalone.test.ts
+++ b/runtimes/runtimes/standalone.test.ts
@@ -13,7 +13,6 @@ import * as authModule from './auth/auth'
 import * as encryptedChatModule from './chat/encryptedChat'
 import * as baseChatModule from './chat/baseChat'
 import { pathToFileURL } from 'url'
-import * as loggingModule from '../runtimes/util/loggingUtil'
 import { LogLevel } from '../runtimes/util/loggingUtil'
 
 describe('standalone', () => {
@@ -40,7 +39,6 @@ describe('standalone', () => {
         stubConnection = stubInterface<vscodeLanguageServer.Connection>()
         stubConnection.console = stubInterface<vscodeLanguageServer.RemoteConsole>()
         stubConnection.telemetry = stubInterface<vscodeLanguageServer.Telemetry>()
-        sinon.stub(loggingModule, 'getLoggingUtility').resolves(mockLoggingUtility)
         sinon.stub(vscodeLanguageServer, 'createConnection').returns(stubConnection)
 
         lspRouterStub = stubInterface<lspRouterModule.LspRouter>()
@@ -57,14 +55,14 @@ describe('standalone', () => {
         let chatStub: sinon.SinonStubbedInstance<encryptedChatModule.EncryptedChat> & encryptedChatModule.EncryptedChat
         let baseChatStub: sinon.SinonStubbedInstance<baseChatModule.BaseChat> & baseChatModule.BaseChat
 
-        it('should initialize without encryption when no key is present', async () => {
+        it('should initialize without encryption when no key is present', () => {
             sinon.stub(authEncryptionModule, 'shouldWaitForEncryptionKey').returns(false)
             authStub = stubInterface<authModule.Auth>()
             sinon.stub(authModule, 'Auth').returns(authStub)
             baseChatStub = stubInterface<baseChatModule.BaseChat>()
             sinon.stub(baseChatModule, 'BaseChat').returns(baseChatStub)
 
-            await standalone(props)
+            standalone(props)
 
             sinon.assert.calledWithExactly(authModule.Auth as unknown as sinon.SinonStub, stubConnection)
             sinon.assert.calledWithExactly(
@@ -72,7 +70,7 @@ describe('standalone', () => {
                 'Runtime: Initializing runtime without encryption'
             )
             sinon.assert.calledWithExactly(baseChatModule.BaseChat as unknown as sinon.SinonStub, stubConnection)
-            sinon.assert.calledOnce(lspRouterStub.servers.push as sinon.SinonStub)
+            sinon.assert.calledTwice(lspRouterStub.servers.push as sinon.SinonStub)
             sinon.assert.calledOnce(stubConnection.listen)
         })
 
@@ -113,8 +111,7 @@ describe('standalone', () => {
                 encryptionInitialization.key,
                 encryptionInitialization.mode
             )
-            await setTimeout(() => Promise.resolve(), 1000)
-            sinon.assert.calledOnce(lspRouterStub.servers.push as sinon.SinonStub)
+            sinon.assert.calledTwice(lspRouterStub.servers.push as sinon.SinonStub)
             sinon.assert.calledOnce(stubConnection.listen)
         })
     })
@@ -122,8 +119,8 @@ describe('standalone', () => {
     describe('features', () => {
         let features: Features
 
-        beforeEach(async () => {
-            await standalone(props)
+        beforeEach(() => {
+            standalone(props)
             features = stubServer.getCall(0).args[0]
         })
 
--- a/runtimes/runtimes/standalone.test.ts
+++ b/runtimes/runtimes/standalone.test.ts
@@ -113,9 +113,9 @@ describe('standalone', () => {
                 encryptionInitialization.key,
                 encryptionInitialization.mode
             )
-            // TODO: The below one fails as it tries to test async inside async, need to check back later
-            // sinon.assert.calledOnce(lspRouterStub.servers.push as sinon.SinonStub)
-            // sinon.assert.calledOnce(stubConnection.listen)
+            await setTimeout(() => Promise.resolve(), 1000)
+            sinon.assert.calledOnce(lspRouterStub.servers.push as sinon.SinonStub)
+            sinon.assert.calledOnce(stubConnection.listen)
         })
     })
 
--- a/runtimes/runtimes/standalone.test.ts
+++ b/runtimes/runtimes/standalone.test.ts
@@ -13,12 +13,22 @@ import * as authModule from './auth/auth'
 import * as encryptedChatModule from './chat/encryptedChat'
 import * as baseChatModule from './chat/baseChat'
 import { pathToFileURL } from 'url'
+import * as loggingModule from '../runtimes/util/loggingUtil'
+import { LogLevel } from '../runtimes/util/loggingUtil'
 
 describe('standalone', () => {
     let stubServer: sinon.SinonStub
     let props: RuntimeProps
     let stubConnection: sinon.SinonStubbedInstance<vscodeLanguageServer.Connection> & vscodeLanguageServer.Connection
     let lspRouterStub: sinon.SinonStubbedInstance<lspRouterModule.LspRouter> & lspRouterModule.LspRouter
+    const mockLoggingUtility = {
+        debug: sinon.stub(),
+        error: sinon.stub(),
+        info: sinon.stub(),
+        warn: sinon.stub(),
+        log: sinon.stub(),
+        level: 'info' as LogLevel,
+    }
 
     beforeEach(() => {
         stubServer = sinon.stub()
@@ -30,6 +40,7 @@ describe('standalone', () => {
         stubConnection = stubInterface<vscodeLanguageServer.Connection>()
         stubConnection.console = stubInterface<vscodeLanguageServer.RemoteConsole>()
         stubConnection.telemetry = stubInterface<vscodeLanguageServer.Telemetry>()
+        sinon.stub(loggingModule, 'getLoggingUtility').resolves(mockLoggingUtility)
         sinon.stub(vscodeLanguageServer, 'createConnection').returns(stubConnection)
 
         lspRouterStub = stubInterface<lspRouterModule.LspRouter>()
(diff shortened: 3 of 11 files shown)


runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+    log: 3,
+    debug: 4,
+}
+
+export const isLogLevelEnabled = (l1: LogLevel, l2: LogLevel): boolean => {
+    return logLevelMap[l1] >= logLevelMap[l2]
+}
+
+export const isValidLogLevel = (level: LogLevel): boolean => {
+    return level && typeof level === 'string' && ['fatal', 'error', 'warn', 'info', 'log', 'debug'].includes(level)
```
ege0zcan, line 19: `fatal` is mentioned in the array but it is not defined in LogLevel above, is that intentional?

runtimes/runtimes/standalone.test.ts
```diff
@@ -102,6 +113,7 @@ describe('standalone', () => {
                 encryptionInitialization.key,
                 encryptionInitialization.mode
             )
+            await setTimeout(() => Promise.resolve(), 1000)
```
Jurredr, line 116: Is this 1s timeout here necessary? If so, why?

runtimes/runtimes/util/loggingUtil.ts
pras0131: my bad; earlier i added fatal but removed it later - forgot to remove it from here. Let me do it. Thanks Ege.

Jurredr: First two checks are redundant.

Also cleaner solution IMO:

```typescript
export type LogLevel = 'error' | 'warn' | 'info' | 'log' | 'debug';

const logLevels: Set<LogLevel> = new Set(['error', 'warn', 'info', 'log', 'debug']);

function isLogLevel(level: string): level is LogLevel {
  return logLevels.has(level as LogLevel);
}
```

Jurredr: Or perhaps even better:
```typescript
const logLevels = [ 'error', 'warn', 'info', 'log', 'debug' ] as const;

export type LogLevel = (typeof logLevels)[number];
```

runtimes/runtimes/standalone.test.ts
pras0131: yes for this specific test case - there is an async call inside an async call. If I don't add the timeout, the function after this line is executed later - so to avoid that situation I have added this timeout.

runtimes/runtimes/util/loggingUtil.ts
pras0131: Thanks Jurre, let me check and test on local & make the change. The response from getConfiguration is any though - I just need to cross check if that works fine with (2) approach you mentioned.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+            }
+        },
+        log: message => {
+            if (isLogLevelEnabled(logLevel, 'log')) {
+                lspConnection.console.log(`[${new Date().toISOString()}] ${message}`)
+            }
+        },
+        debug: message => {
+            if (isLogLevelEnabled(logLevel, 'debug')) {
+                lspConnection.console.debug(`[${new Date().toISOString()}] ${message}`)
```
natdurlik, line 63: What do you think about extracting `[${new Date().toISOString()}] ${message}` part to a separate function? We repeat it for each log level.

tanujbhakre:
Not every one would have access to internal tools, could you please remove the SIM link and provide the context?

runtimes/runtimes/util/loggingUtil.ts
pras0131: It was a small string and there are only 5 levels, so that's why I thought looking at it should be clear specially incase of logging what we are logging.

runtimes/runtimes/standalone.ts
```diff
@@ -133,15 +134,9 @@ export const standalone = (props: RuntimeProps) => {
...
+    async function initializeRuntime(encryptionKey?: string) {
         const documents = new TextDocuments(TextDocument)
-
-        // Set up logging over LSP
-        // TODO: set up Logging once implemented
-        const logging: Logging = {
-            log: message => lspConnection.console.info(`[${new Date().toISOString()}] ${message}`),
-        }
-
+        const logging: Logging = await getLoggingUtility(lspConnection)
```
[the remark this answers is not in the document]
pras0131: Initially I also thought to keep it there but added it here because all other logic related to init runtime is kept inside initializeRuntime method.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+        logLevel = await getLogLevel(lspConnection)
+    } catch {
+        logLevel = DEFAULT_LOG_LEVEL
+    }
+
+    const logging: Logging = {
+        level: logLevel,
+        error: message => {
+            if (isLogLevelEnabled(logLevel, 'error')) {
+                lspConnection.console.error(`[${new Date().toISOString()}] ${message}`)
```
viktorsaws, line 43: Do we really need to append date to each log message? Imo, it pollutes original string to much, and should not be necessary - `console` calls handlers can print date as well, which will make this redundant.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const getLoggingUtility = async (lspConnection: Connection): Promise<Logging> => {
+    let logLevel: LogLevel
+    try {
+        logLevel = await getLogLevel(lspConnection)
+    } catch {
+        logLevel = DEFAULT_LOG_LEVEL
+    }
+
+    const logging: Logging = {
```
viktorsaws, line 39: Can we convert this object to Class and instantiate logger from it? Will help to better organize code

runtimes/runtimes/util/loggingUtil.ts
viktorsaws: We can have 1 `.sendToLog` method indeed, that accepts log level and does all needed arguments processing indeed, will be cleaner.

runtimes/runtimes/standalone.ts
```diff
@@ -92,7 +93,7 @@ if (checkAWSConfigFile()) {
  * @param props.servers The list of servers to initialize and run
  * @returns
  */
-export const standalone = (props: RuntimeProps) => {
+export const standalone = async (props: RuntimeProps) => {
```
viktorsaws, line 96: We should avoid changing signature of `standalone` function, as it is public interface for all consuming packages - we have many build configs today, and it is in a way breaking change.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
+import { Logging } from '../../server-interface'
+import { Connection } from 'vscode-languageserver/node'
+
+export type LogLevel = 'error' | 'warn' | 'info' | 'log' | 'debug'
+
+const logLevelMap: Record<LogLevel, number> = {
```
viktorsaws, line 6: Let's align log level enum with standard LSP `MessageType` interface 
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#messageType. We can use it here directly

runtimes/runtimes/standalone.ts
viktorsaws: I see that this change is caused by `getLoggingUtility` requesting configuration from client in `getLogLevel()` with `lspConnection.workspace.getConfiguration`.

I think this is not correct approach, as it breaks LSP specification - it requests something by server from client before server even starts listening on the connection, and before server responds with `Initialize`. Were you able to test that it actually works with compiled server and real client? I suspect this may not work correctly.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize

> Until the server has responded to the initialize request with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult), the client must not send any additional requests or notifications to the server. In addition **the server is not allowed to send any requests or notifications to the client until it has responded with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult)**, with the exception that during the initialize request the server is allowed to send the notifications window/showMessage, window/logMessage and [telemetry/event](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event) as well as the window/showMessageRequest request to the client.


Better solution will be:

* Accept `logLevel` option in [InitializeParams](https://github.com/aws/language-server-runtimes/blob/f6be37f05713ec5d1fc8156c8567f9e948b3894b/runtimes/protocol/lsp.ts#L115-L116).initializationOptions. Set reasonable default value, when not set (`log` or `info`)
* In runtime, subscribe to [workspace/didChangeConfiguration](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration) notification from client, and only then pull and update config with `getLogLevel()`.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const isValidLogLevel = (level: LogLevel): boolean => {
+    return level && typeof level === 'string' && ['error', 'warn', 'info', 'log', 'debug'].includes(level)
+}
+
+const DEFAULT_LOG_LEVEL: LogLevel = 'log'
+
+export const getLogLevel = async (lspConnection: Connection): Promise<LogLevel> => {
+    const logLevelConfig = await lspConnection.workspace.getConfiguration({
+        section: 'logLevel',
```
viktorsaws, line 26: Let prepend section with our own prefix to avoid accidental collision with other configs. We can use `aws.logLevel` here to scope this down.

viktorsaws (changes requested):

runtimes/runtimes/standalone.ts
pras0131: I added the test cases but did not try out testing with the real client though. 

>it requests something by server from client before server even starts listening on the connection, and before server responds with Initialize
I see lspConnection.listen is at the end of the file. 

If I go ahead with **workspace/didChangeConfiguration**, then a part of the runtime will be initialized later while the other components would have been initialized early. Or you are suggesting to start with initializing logging as default log/info log level and then listen to configuration update and then update the logging level to a new value?

runtimes/runtimes/util/loggingUtil.ts
pras0131: I kept it to keep it aligned with what we had earlier.

runtimes/runtimes/util/loggingUtil.ts
pras0131: I can do that but then in that function there will be if/else based on log level received in the argument.

runtimes/runtimes/standalone.ts
pras0131: I would prefer to go with workspace/didChangeConfiguration (if my understanding is correct as described in the above comment) over InitializeParams as it's more dynamic and we don't need to update the structure of initializeOptions as well.

pras0131: A doubt though: I think with didChangeConfiguration one as well, we already return the logging to all servers before while iterating them. Don't know if updating logging later would help here.

viktorsaws: We need both to support this feature fully: optional initial `logLevel` value passed in `initializationOptions` and subscription to `workspace/didChangeConfiguration`. 
Without `initializationOptions` server will always default to some value, while clients may be able to preset `logLevel` from the very start

runtimes/runtimes/standalone.test.ts
```diff
@@ -13,12 +13,22 @@ import * as authModule from './auth/auth'
...
     let stubServer: sinon.SinonStub
     let props: RuntimeProps
     let stubConnection: sinon.SinonStubbedInstance<vscodeLanguageServer.Connection> & vscodeLanguageServer.Connection
     let lspRouterStub: sinon.SinonStubbedInstance<lspRouterModule.LspRouter> & lspRouterModule.LspRouter
+    const mockLoggingUtility = {
+        debug: sinon.stub(),
+        error: sinon.stub(),
+        info: sinon.stub(),
+        warn: sinon.stub(),
+        log: sinon.stub(),
```
justinmk3, line 29: why stub this stuff? just assert the actual log messages.

runtimes/server-interface/logging.ts
```diff
@@ -1,7 +1,13 @@
+import { LogLevel } from '../runtimes/util/loggingUtil'
+
 /**
  * The logging feature interface
  */
 export type Logging = {
-    // TODO: Implement log levels
+    level: LogLevel
```
volodkevych, line 7: Why is it part of server interface? Is it expected to be used by servers?

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,58 @@
...
+            this.lspConnection.console.info(`[${new Date().toISOString()}] ${message}`)
+        }
+    }
+    log = (message: string) => {
+        if (isLogLevelEnabled(this.level, 'log')) {
+            this.lspConnection.console.log(`[${new Date().toISOString()}] ${message}`)
+        }
+    }
+    debug = (message: string) => {
+        if (isLogLevelEnabled(this.level, 'debug')) {
```
viktorsaws, line 54: nit: can reduce duplication, since almost all the same code is called 5 times?

runtimes/runtimes/lsp/router/loggingServer.ts
```diff
@@ -0,0 +1,44 @@
...
+        this.lspServer = new LspServer(this.lspConnection, this.encoding, this.logger)
+        this.lspServer.setInitializeHandler(async (params: InitializeParams): Promise<InitializeResult> => {
+            this.updateLoggingLevel(params.initializationOptions?.logLevel ?? ('log' as LogLevel))
+            return {
+                capabilities: {},
+            }
+        })
+        this.lspServer.setDidChangeConfigurationHandler(async params => {
+            const logLevelConfig = await lspConnection.workspace.getConfiguration({
+                section: 'aws.logLevel',
```
viktorsaws, line 25: We need to add this LSP configuration details and supported values to documentation, to communicate it to destinations https://github.com/aws/language-server-runtimes/blob/main/runtimes/README.md#logging

runtimes/runtimes/util/loggingUtil.ts
pras0131: The method inside each of these is different..Depending what we want to log..sometimes lspconnection.console.log/debug/info etc

viktorsaws: we can uses strings to access properties, something like this `this.lspConnection['console']()`

runtimes/server-interface/logging.ts
```diff
@@ -1,7 +1,13 @@
...
+    error: (message: string) => void
+    warn: (message: string) => void
+    info: (message: string) => void
     log: (message: string) => void
+    debug: (message: string) => void
```
volodkevych, line 12: nit: we could add something as `...args: any` for convernience if we are already improving this interface

runtimes/server-interface/logging.ts
volodkevych: We define method per each log level below - it should be enough for the server developer to know that we support log levels.

runtimes/runtimes/util/loggingUtil.ts
natdurlik: we can pass different method (lspconnection.console.log/debug/info) as an argument

runtimes/runtimes/util/loggingUtil.test.ts
```diff
@@ -0,0 +1,78 @@
...
+                { level1: 'error', level2: 'error', expected: true },
+            ]
+            testCases.forEach(({ level1, level2, expected }) => {
+                assert.strictEqual(expected, isLogLevelEnabled(level1 as LogLevel, level2 as LogLevel))
+            })
+        })
+    })
+
+    describe('LoggingImplementation', () => {
+        it('should test implementation of Logging', async () => {
```
volodkevych, line 58: "should check logging method of configured log level are called" would be a better name for the test

runtimes/runtimes/lsp/router/loggingServer.ts
```diff
@@ -0,0 +1,44 @@
...
+            })
+            if (isValidLogLevel(logLevelConfig)) {
```
volodkevych, line 27: Should we log error if log level is invalid?

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,58 @@
...
+    log: MessageType.Log,
+    debug: MessageType.Debug,
+}
+
+export const isLogLevelEnabled = (l1: LogLevel, l2: LogLevel): boolean => {
+    return logLevelMap[l1] >= logLevelMap[l2]
+}
+
+export const isValidLogLevel = (level: LogLevel): boolean => {
+    return ['error', 'warn', 'info', 'log', 'debug'].includes(level)
```
volodkevych, line 19: You are copy-pasting these values throughout the code - it's much better to use `MessageType` enum.

runtimes/runtimes/lsp/router/loggingServer.ts
```diff
@@ -0,0 +1,44 @@
...
+                this.updateLoggingLevel(logLevelConfig as LogLevel)
+            }
+        })
+    }
+
+    private updateLoggingLevel(logLevel: LogLevel) {
+        this.logger.level = logLevel
```
volodkevych, line 34: Ideally, I would also log that the logging level changed.

runtimes/runtimes/lsp/router/loggingServer.ts
```diff
@@ -0,0 +1,44 @@
...
+    private lspServer: LspServer
+
+    constructor(
+        private lspConnection: Connection,
+        private encoding: Encoding
+    ) {
+        this.logger = new LoggingImplementation(DEFAULT_LOG_LEVEL as LogLevel, this.lspConnection)
+        this.lspServer = new LspServer(this.lspConnection, this.encoding, this.logger)
+        this.lspServer.setInitializeHandler(async (params: InitializeParams): Promise<InitializeResult> => {
+            this.updateLoggingLevel(params.initializationOptions?.logLevel ?? ('log' as LogLevel))
```
volodkevych, line 18: Do we have `logLevel` defined in `initializationOptions` in protocol?

pras0131: we don't need to define the same explicitly inside `initializationOptions` as it contains a map of string to any. Reference: [link](https://github.com/aws/language-server-runtimes/blob/b8e3adeff5a8396afc40062439b9541512a0d29d/runtimes/protocol/lsp.ts#L116)

runtimes/runtimes/lsp/router/loggingServer.ts
pras0131: yes, let me add that.

runtimes/runtimes/lsp/router/loggingServer.ts
volodkevych: It's good to have it explicitly listed and/or documented. Otherwise, how will clients know what is possible?

runtimes/runtimes/lsp/router/loggingServer.ts
pras0131: Not doing it right now as currently no destination is using this aws.logLevel value, so there will be false logs.

runtimes/server-interface/logging.ts
pras0131: I could add but logger which we are using lspConnection.console.log/error/warn/info/debug does not support passing extra arguments, they accept a single string argument only.

runtimes/runtimes/lsp/router/loggingServer.ts
pras0131: I'll update with the details under README.md

runtimes/server-interface/logging.ts
volodkevych: What you describe is a part of runtime (LoggingServer is inside of runtimes, it's not real server). Server interface that you are extending here exposes `level` to the servers/capabilities that are implemented by service teams. It gives them possibility to read and also modify the logging level, which does not make sense.

runtimes/runtimes/lsp/router/loggingServer.ts
```diff
@@ -0,0 +1,45 @@
...
+            })
+            if (isValidLogLevel(logLevelConfig)) {
+                this.updateLoggingLevel(logLevelConfig as LogLevel)
+            }
+        })
+    }
+
+    private updateLoggingLevel(logLevel: LogLevel) {
+        this.logger.level = logLevel
+        this.logger.info(`Logging level changed to ${logLevel}`)
```
volodkevych, line 35: We should log this only if value really changed. I think it's possible that other config value changed that triggers `setDidChangeConfigurationHandler` and then we log this, which could be misleading.

runtimes/runtimes/util/loggingUtil.ts
```diff
@@ -0,0 +1,44 @@
...
+        this.lspConnection = connection
+    }
+
+    sendToLog(logLevel: LogLevel, message: string): void {
+        if (isLogLevelEnabled(this.level, logLevel)) {
+            this.lspConnection.console[logLevel](`[${new Date().toISOString()}] ${message}`)
+        }
+    }
+
+    error = (message: string) => this.sendToLog('error', message)
```
volodkevych, line 39: nit: we still have these log level strings copied few times, especially in tests. Could be possibly improved with enums.

volodkevych (approved):

runtimes/README.md
```diff
@@ -136,11 +136,16 @@ Design TBD
...
-TBD
+The runtimes by default supports the logging feature, allowing servers to output logs varied by different log levels. 
 
 #### Feature Specification
 
-TBD
+The log level to be set in the runtime for logging can be decided by the client as part of initialize request, which is the first
+request initiated by client to the LSP server. The log level can be updated dynamically even after server start by triggering `workspace/didChangeConfiguration` 
+notification from the client - which prompts the runtime to fetch the new log level to be set from the client through LSP `getConfigurations` request. 
+1. Params in the initialize request: `InitializeParams.initializationOptions.logLevel`
```
viktorsaws, line 146: `logLevel` also needs to be added to protocol for `initializationOptions` here https://github.com/aws/language-server-runtimes/blob/main/runtimes/protocol/lsp.ts#L115-L118, to keep it explicit

runtimes/runtimes/lsp/router/loggingServer.ts
viktorsaws: As I mentioned in another comment, I think we need to explicitly add `logLevel` to protocol, since that's how we define contract with destination clients. It has to be added explicitly, as otherwise we may accidentally break protocol without noticing

viktorsaws (approved):

PR #292, opened 2025-01-03
fix(types): fixed publish package contents

dogusata:
## Problem
Even though the previous release fixed the `main` path targeting to the correct `./out` folder, it contains `.ts` files (not the `.d.ts`) which shouldn't be in the package. Also, some files like NOTICE, SECURITY etc. are missing.

## Solution
Updated the `.npmignore` and fixed the copy files on `prepub`.

## License

### DIFF
--- a/types/package.json
+++ b/types/package.json
@@ -6,7 +6,7 @@
     "scripts": {
         "clean": "rm -rf out/",
         "compile": "tsc --build",
-        "prepub:copyFiles": "shx cp ../.npmignore CHANGELOG.md ../LICENSE ../NOTICE README.md ../SECURITY.md package.json out/",
+        "prepub:copyFiles": "shx cp ../LICENSE ../NOTICE ../SECURITY.md .",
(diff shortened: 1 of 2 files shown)


types/package.json
```diff
@@ -6,7 +6,7 @@
     "scripts": {
         "clean": "rm -rf out/",
         "compile": "tsc --build",
-        "prepub:copyFiles": "shx cp ../.npmignore CHANGELOG.md ../LICENSE ../NOTICE README.md ../SECURITY.md package.json out/",
```
ege0zcan, line 9: Are we removing CHANGELOG.md and README.md from here because they are published inside the `out` folder?

dogusata: No, they are in the main root folder already, it was being done to move them inside the .out folder with the previous broken version. We're just adding NOTICE, LICENSE and SECURITY files.

ege0zcan: Oh ok I see now, that makes sense thanks

ege0zcan (approved):

PR #295, opened 2025-01-07
feat: add env variables getter in runtime

francescoopiccoli:
## Problem
We want to be able to inject environment variables via runtime
## Solution
Added a environment variables getter in `Runtime`, for the webworker returns undefined as `process`object  is not defined for webworker. 

## License

### DIFF
--- a/runtimes/server-interface/runtime.ts
+++ b/runtimes/server-interface/runtime.ts
@@ -21,9 +21,9 @@ export interface Runtime {
     platform: Platform
 
     /**
-     * Get the value of an environment variable.
-     * @param key The name of the environment variable.
-     * @returns The value of the environment variable or undefined if the environment variable is not set.
+     * Get a runtime configuration value.
+     * @param key The configuration key to retrieve.
+     * @returns The configuration value or undefined if the key is not set.
      */
-    getEnvironmentConfiguration(key: string): string | undefined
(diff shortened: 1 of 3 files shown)


runtimes/server-interface/runtime.ts
```diff
@@ -19,4 +19,11 @@ export interface Runtime {
      * Set to NodeJS.Platform for standalone and 'browser' for webworker runtime.
      */
     platform: Platform
+
+    /**
+     * Get the value of an environment variable.
```
imykhai, line 24: Calling this environment variable implies that it's passed to the process. 
But non-standalone runtimes aren't run in the processes. It's better to find a better more general name/description. 
I think, the `configuration` part in the `getEnvironmentConfiguration` makes sense, but then in the standalone runtime we should document that it gets the configurations from the environment variables, while `webworker` or any other runtimes will get the configuration from other sources.

francescoopiccoli: good point, what about this?
```
     /**
     * Get a runtime configuration value.
     * @param key The configuration key to retrieve.
     * @returns The configuration value or undefined if not set.
     */
    getConfiguration(key: string): string | undefined
```

imykhai: looks good!

PR #299, opened 2025-01-15
feat: adding support for readFileSync method in workspace.fs

pras0131:
Adding support for readFileSync method in workspace.fs feature of runtimes.

## License

### DIFF
--- a/runtimes/runtimes/webworker.ts
+++ b/runtimes/runtimes/webworker.ts
@@ -48,7 +48,6 @@ import {
 import { IdentityManagement } from '../server-interface/identity-management'
 import { Encoding, WebBase64Encoding } from './encoding'
 import { LoggingServer } from './lsp/router/loggingServer'
-import { Buffer } from 'node:buffer'
 
 declare const self: WindowOrWorkerGlobalScope
 
--- a/runtimes/runtimes/webworker.ts
+++ b/runtimes/runtimes/webworker.ts
@@ -86,7 +86,7 @@ export const webworker = (props: RuntimeProps) => {
             writeFile: (_path, _data) => Promise.resolve(),
             appendFile: (_path, _data) => Promise.resolve(),
             mkdir: (_path, _options?) => Promise.resolve(''),
-            readFileSync: _path => Buffer.alloc(0),
+            readFileSync: (_path, _options?) => '',
         },
     }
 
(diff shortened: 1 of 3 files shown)


runtimes/runtimes/webworker.ts
```diff
@@ -48,6 +48,7 @@ import {
 import { IdentityManagement } from '../server-interface/identity-management'
 import { Encoding, WebBase64Encoding } from './encoding'
 import { LoggingServer } from './lsp/router/loggingServer'
+import { Buffer } from 'node:buffer'
```
natdurlik, line 51: Can we import it if we don't run on node environment here?

natdurlik: The workspace interface should be environment agnostic.

pras0131: I think the buffer library got imported from here by mistake. Let me fix it.

francescoopiccoli: I think importing this would break the bundling of the webworker as it happened with 'process'. I am wondering whether we handle the return value as a string as we do for `readFile`?

pras0131: @francescoopiccoli  I removed the import.

runtimes/server-interface/workspace.ts
```diff
@@ -49,5 +49,6 @@ export type Workspace = {
         writeFile: (path: string, data: string) => Promise<void>
         appendFile: (path: string, data: string) => Promise<void>
         mkdir: (path: string, options?: { recursive?: boolean }) => Promise<string | undefined>
+        readFileSync: (path: string) => Buffer
```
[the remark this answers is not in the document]
pras0131: Earlier I didn't see there is another variant with the string return available as well in the fs library. Let me make the return type as string.

runtimes/runtimes/webworker.ts
pras0131: nice catch. There might be a runtime issue. Changing the signature to return the value of string instead.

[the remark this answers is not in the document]
pras0131: fixed

natdurlik: Could you also add a doc comment in workspace for `readFileSync` function similar to `readFile` that points out  that we set a default encoding? So users won't be confused that our function behaves differently from the Node.js version?

natdurlik (approved):

ege0zcan (commented):
Changes look good to me but I would consider this a new feature rather than a fix

pras0131:
> Changes look good to me but I would consider this a new feature rather than a fix

done, changed it to a feature

PR #306, opened 2025-01-23
fix: handling of aws object of initializationOptions and initialize error handling

francescoopiccoli:
## Problem
Prevent errors due to aws object being undefined and improve error handling in initialize method
 
## Solution
Access aws object safely via with optional chaining and wrap the initialize method in a try/catch block logging error info

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/lspServer.ts
+++ b/runtimes/runtimes/lsp/router/lspServer.ts
@@ -88,7 +88,7 @@ export class LspServer {
         token: CancellationToken
     ): Promise<PartialInitializeResult | ResponseError<InitializeError> | undefined> => {
         this.clientSupportsNotifications =
-            params.initializationOptions?.aws.awsClientCapabilities?.window?.notifications
+                params.initializationOptions?.aws?.awsClientCapabilities?.window?.notifications
 
         if (!this.initializeHandler) {
             return
--- a/runtimes/runtimes/lsp/router/lspServer.ts
+++ b/runtimes/runtimes/lsp/router/lspServer.ts
@@ -87,22 +87,29 @@ export class LspServer {
         params: InitializeParams,
         token: CancellationToken
     ): Promise<PartialInitializeResult | ResponseError<InitializeError> | undefined> => {
-        this.clientSupportsNotifications =
+        try {
+            this.clientSupportsNotifications =
                 params.initializationOptions?.aws?.awsClientCapabilities?.window?.notifications
 
-        if (!this.initializeHandler) {
+            if (!this.initializeHandler) {
+                return
+            }
+
+            const initializeResult = await asPromise(this.initializeHandler(params, token))
+            if (!(initializeResult instanceof ResponseError)) {
+                this.initializeResult = initializeResult
+                if (initializeResult?.serverInfo) {
+                    this.notificationRouter = new RouterByServerName(initializeResult.serverInfo.name, this.encoding)
+                }
+            }
+
+            return initializeResult
+        } catch (error) {
+            this.lspConnection.console.log(
+                `Error in initialize handler: "${error}",\nwith initialization options: ${JSON.stringify(params.initializationOptions)}`
+            )
             return
         }
-
-        const initializeResult = await asPromise(this.initializeHandler(params, token))
-        if (!(initializeResult instanceof ResponseError)) {
-            this.initializeResult = initializeResult
-            if (initializeResult?.serverInfo) {
-                this.notificationRouter = new RouterByServerName(initializeResult.serverInfo.name, this.encoding)
-            }
-        }
-
-        return initializeResult
     }
 
     public tryExecuteCommand = async (
--- a/runtimes/runtimes/util/serverDataDirPath.ts
+++ b/runtimes/runtimes/util/serverDataDirPath.ts
@@ -3,7 +3,7 @@ import * as os from 'os'
 import { InitializeParams } from '../../protocol'
 
 export function getServerDataDirPath(serverName: string, initializeParams: InitializeParams | undefined): string {
-    const clientSpecifiedLocation = initializeParams?.initializationOptions?.aws.clientDataFolder
+    const clientSpecifiedLocation = initializeParams?.initializationOptions?.aws?.clientDataFolder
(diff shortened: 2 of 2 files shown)


imykhai (approved):

runtimes/runtimes/lsp/router/lspServer.ts
```diff
@@ -87,22 +87,29 @@ export class LspServer {
...
             }
-        }
 
-        return initializeResult
+            return initializeResult
+        } catch (error) {
+            this.lspConnection.console.log(
+                `Error in initialize handler: "${error}",\nwith initialization options: ${JSON.stringify(params.initializationOptions)}`
+            )
+            return
```
awschristou, line 111: Would we still want to throw the error? This would inform clients that the server cannot start up. (We would want to return the formatted message above as the error state message). 

To put it another way: Is there a chance that by swallowing the error, the server will be in an unstable state (with unpredictable behavior)?

runtimes/runtimes/lsp/router/lspServer.ts
```diff
@@ -87,22 +87,29 @@ export class LspServer {
...
+                if (initializeResult?.serverInfo) {
+                    this.notificationRouter = new RouterByServerName(initializeResult.serverInfo.name, this.encoding)
+                }
             }
-        }
 
-        return initializeResult
+            return initializeResult
+        } catch (error) {
+            this.lspConnection.console.log(
```
awschristou, line 108: Can you confirm that this message would be relayed to clients? I'm concerned that without a successful initialize handshake, this message may not be seen or consumed anywhere useful. We may have to throw an error that contains this text in order for clients to receive it (but this depends on my other comment in this PR)

runtimes/runtimes/lsp/router/lspServer.ts
```diff
@@ -87,22 +87,29 @@ export class LspServer {
...
+                this.initializeResult = initializeResult
+                if (initializeResult?.serverInfo) {
+                    this.notificationRouter = new RouterByServerName(initializeResult.serverInfo.name, this.encoding)
+                }
             }
-        }
 
-        return initializeResult
+            return initializeResult
+        } catch (error) {
```
volodkevych, line 107: What's the reason to try/catch here?

volodkevych: I think with this change LSP server will start even if Flare servers inside failed to initialize, right? How will client know about that?

runtimes/runtimes/lsp/router/lspServer.ts
```diff
@@ -87,22 +87,29 @@ export class LspServer {
         params: InitializeParams,
         token: CancellationToken
     ): Promise<PartialInitializeResult | ResponseError<InitializeError> | undefined> => {
-        this.clientSupportsNotifications =
-            params.initializationOptions?.aws.awsClientCapabilities?.window?.notifications
+        try {
+            this.clientSupportsNotifications =
+                params.initializationOptions?.aws?.awsClientCapabilities?.window?.notifications
```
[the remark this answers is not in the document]
imykhai: All of the subfields of the `aws` are optional, why do we require `aws` be mandatory? 
In this particular case client doesn't pass the aws->window->notifications so it's a fair assumption they don't support notifications, why do we need to fail initialization in this case?

runtimes/runtimes/lsp/router/lspServer.ts
imykhai: yes the primary reason was for logging but indeed, let's re-throw the exception.

runtimes/runtimes/lsp/router/lspServer.ts
imykhai: yes, [the specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) makes exception to logs and telemetry:

Until the server has responded to the initialize request with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult), the client must not send any additional requests or notifications to the server. In addition the server is not allowed to send any requests or notifications to the client until it has responded with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult), with the exception that during the initialize request the server is allowed to send the notifications window/showMessage, window/logMessage and [telemetry/event](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#telemetry_event) as well as the window/showMessageRequest request to the client.

runtimes/runtimes/lsp/router/lspServer.ts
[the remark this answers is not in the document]
volodkevych: If `aws` is optional, the protocol has to set proper expectations.

runtimes/runtimes/lsp/router/lspServer.ts
francescoopiccoli: fix in this PR: https://github.com/aws/language-server-runtimes/pull/307

runtimes/runtimes/lsp/router/lspServer.ts
francescoopiccoli: fix in this PR: https://github.com/aws/language-server-runtimes/pull/307

runtimes/runtimes/lsp/router/lspServer.ts
francescoopiccoli: yes I tested manually, I can confirm this as well

runtimes/runtimes/lsp/router/lspServer.ts
justinmk3, line 92: There is no reason for this line to be in a `try` block. The try-catch should be as minimal as possible, to make it clear what's being guarded, and it would also have helped with reviewing this PR.

francescoopiccoli: Errata: sorry for the confusion, we had some back and forth with the team, and realized that in this PR we were trying to pursue two contrasting objectives, mitigating and logging the issue. the latest changes and comments, in (#307) explain. We need to keep this line inside this block, (and remove optional chaining) to log details about such failures.

PR #312, opened 2025-01-31
feat: add operational telemetry interface

natdurlik:
## Problem
We need to be able the collect and export operational telemetry data standalone runtimes in a consistent and secure manner.

## Solution
Defining telemetry interface and metric schema and implementing with opentelemetry sdk.

Check README.md for examples.

## License


runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -0,0 +1,22 @@
+export type OperationalMetric = {
+    name: string
+    value: number
+    timestamp: number
+    atrributes: { string: any }[]
+    metrics: {}
```
volodkevych, line 6: what is this field for?

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -0,0 +1,22 @@
...
+    server: {
+        name: string
+        version: string
+    }
+    clientInfo: {
```
volodkevych, line 11: I think we will need all the fields from `clientInfo` that we get from destination.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+import opentelemetry, { Attributes } from '@opentelemetry/api'
+import { NodeSDK } from '@opentelemetry/sdk-node'
+import { Resource } from '@opentelemetry/resources'
+import { ATTR_SERVICE_NAME, ATTR_SERVICE_VERSION } from '@opentelemetry/semantic-conventions/*'
+import { randomUUID } from 'crypto'
+
+export class OperationalTelemetryService implements OperationalTelemetry {
+    private customResource: Record<string, any> = {}
+    private static instance: OperationalTelemetryService
+    private SCOPE_NAME = 'language-server-runtimes'
```
volodkevych, line 13: This needs to scale to servers in the future.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    static getInstance(): OperationalTelemetryService {
+        if (!OperationalTelemetryService.instance) {
+            OperationalTelemetryService.instance = new OperationalTelemetryService()
+        }
+        return OperationalTelemetryService.instance
+    }
+
+    initialize(serviceName: string, serviceVersion: string): void {
```
volodkevych, line 49: This signature gives false understanding that you can initialize this service for different name and version, however everything after the 1st call is just ignored.

volodkevych: Do we need `getInstance` and ` initialize` as a separate functions on the interface?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+        if (this.initialized) {
+            return
+        }
+        this.initialized = true
+
+        const poolId = ''
+        const region = ''
+        const endpoint = ''
```
volodkevych, line 57: How will those values be set?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+                [ATTR_SERVICE_VERSION]: serviceVersion,
+                sessionId: randomUUID(),
+            }),
+            metricReader: metricReader,
+        })
+
+        sdk.start()
+    }
+
+    getResource(): Record<string, any> {
```
volodkevych, line 78: No experience with open telemetry, but could resources be extended using `resourceDetectors` somehow. I would be cleaner.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    private constructor() {}
+
+    reportCounterMetric(counterName: string, value = 1, attributes?: Record<string, any>): void {
+        if (!this.initialized) {
+            // todo log diag error
+            return
+        }
+
+        const meter = opentelemetry.metrics.getMeter(this.SCOPE_NAME)
```
volodkevych, line 24: Do we need to call `getMeter` in every call?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,85 @@
...
+export class OperationalTelemetryService implements OperationalTelemetry {
+    private customResource: Record<string, any> = {}
+    private static instance: OperationalTelemetryService
+    private SCOPE_NAME = 'language-server-runtimes'
+    private initialized = false
+
+    private constructor() {}
+
+    reportCounterMetric(counterName: string, value = 1, attributes?: Record<string, any>): void {
+        if (!this.initialized) {
```
volodkevych, line 19: Can be avoided if we merge getInstance and  initialize.

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
```diff
@@ -0,0 +1,172 @@
...
+        this.isShutdown = true
+        return Promise.resolve()
+    }
+
+    async refreshCognitoCredentials(region: string, poolId: string): Promise<Credentials> {
+        if (this.credentials && this.credentialsLastFetched) {
+            const now = new Date()
+            const timeSinceLastFetch = now.getTime() - this.credentialsLastFetched.getTime()
+
+            if (timeSinceLastFetch < this.CREDENTIALS_EXPIRATION_TIME_MS - this.CREDENTIALS_BUFFER_TIME_MS) {
```
volodkevych, line 68: Shouldn't it be ">"? Can we have tests for such kind of functionality added?

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
```diff
@@ -0,0 +1,172 @@
...
+                version: resource.attributes['service.version'] as string,
+            },
+            clientInfo: {
+                name: this.telemetryService.getResource()['clientInfo.name'] as string,
+            },
+        }
+    }
+
+    async sendOperationalMetrics(metrics: OperationalMetric[]) {
+        for (const metric of metrics) {
```
volodkevych, line 151: Should we batch the events sent over http?

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
```diff
@@ -0,0 +1,172 @@
...
+            .map((scopeMetrics: ScopeMetrics) => {
+                return scopeMetrics.metrics.map((metric: MetricData) => {
+                    return this.transform(metric, metrics.resource)
+                })
+            })
+            .flat()
+    }
+
+    transform(metric: MetricData, resource: Resource): OperationalMetric {
+        const dataPoint = metric.dataPoints[0]
```
volodkevych, line 133: Why only the 1st datapoint? What if there are none?

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
```diff
@@ -0,0 +1,172 @@
...
+                    method: 'POST',
+                    url: this.endpoint,
+                    data: body,
+                    headers: signedRequest.headers,
+                })
+
+                // todo proper logging
+                console.log('Status Code:', response.status)
+                console.log('Response:', response.data)
+            } catch (e) {
```
volodkevych, line 167: Retries for some cases may be needed. How will be know about these failures?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: I've included `initialize` in the operational telemetry interface. This method should be responsible for setting up all instances needed for collecting and sending metrics. I didn't add `getInstance` to the interface because it is up to implementation how the telemetry service is created. 

I can move these parameters to `getInstance` method, what do you think?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: Currently hardcoded in this file. We need to think of a better solution, ideally by ingesting different endpoints for test bundles.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: Could you provide more details about `resourceDetectors`? I plan to use this `customResource` record mainly for adding `clientInfo` to our collected metrics. I can't add `clientInfo` to the opentelemetry resource in the `initialize` method, because it's not available at the beginning of the standalone runtimes function.

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
natdurlik: I will add some unit tests after the logic is done. I tested it manually.

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
natdurlik: still WIP

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
```diff
@@ -0,0 +1,172 @@
...
+            })
+            .flat()
+    }
+
+    transform(metric: MetricData, resource: Resource): OperationalMetric {
+        const dataPoint = metric.dataPoints[0]
+        return {
+            name: metric.descriptor.name,
+            value: dataPoint.value as number,
+            timestamp: dataPoint.endTime[0],
```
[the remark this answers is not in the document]
natdurlik: It's a high resolution time represented by two numbers in an array: [UNIX Epoch time in seconds, partial second in nanoseconds]. I assumed timestamp in seconds would be enough in our case, what do you think?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: You're right, I will include it in the metrics schema.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
volodkevych: Do you target delayed/controlled call to set up all instances for metrics? Why do you want `initialize` in the interface explicitly?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
volodkevych: https://opentelemetry.io/docs/languages/js/resources/

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
volodkevych: Any recommendation for how to use it? It looks like something static that can be defined once to me.

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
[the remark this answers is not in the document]
volodkevych: Ok, should be enough to start with.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: We can pass a promise that will detect certain attributes later (specifically `clientInfo`). However, we won't be able to retrieve any attributes from the resource before this promise resolves. In my opinion, we should keep `clientInfo` optional in our metrics and add it via our custom resources later if available, since there might be some error metrics that occur before initialization. What do you think?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: Okay, we can move the `initialize` logic to the `getInstance` method. Wouldn't the signature of the new `getInstance` method then give a false impression that we could have multiple instances?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: > It’s generally recommended to call getMeter in your app when you need it rather than exporting the meter instance to the rest of your app. This helps avoid trickier application load issues when other required dependencies are involved.
from https://opentelemetry.io/docs/languages/js/instrumentation/

natdurlik: In the future, if we want to have different scopes, we would probably need to obtain the meter in the method anyway.

runtimes/runtimes/operational-telemetry/aws-metric-exporter.ts
natdurlik: We will log failures on in the standalone runtime, and we will also see requests on the api gateway side.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,126 @@
...
+import { randomUUID } from 'crypto'
+import { RemoteConsole } from 'vscode-languageserver'
+import { AwsCognitoApiGatewaySender } from './aws-cognito-gateway-sender'
+import { AwsSpanExporter } from './aws-spans-exporter'
+import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base'
+
+export class OperationalTelemetryService implements OperationalTelemetry {
+    private customAttributes: Record<string, any> = {}
+    private static instance: OperationalTelemetryService
+    private readonly SCOPE_NAME = 'language-server-runtimes'
```
volodkevych, line 17: Scope should be configurable, not hardcoded. This service should be reusable with other scopes.

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
```diff
@@ -0,0 +1,121 @@
...
+    private spanToOperationalEvent(span: ReadableSpan): CaughtErrorEvent | ServerCrashEvent {
+        const unixEpochSeconds = span.endTime[0]
+        if (span.name === 'CaughtErrorEvent') {
+            return {
+                name: span.name,
+                timestamp: unixEpochSeconds,
+                errorType: span.attributes['errorType'] as string,
+            } as CaughtErrorEvent
+        }
+        if (span.name === 'ServerCrashEvent') {
```
volodkevych, line 112: Can we avoid extending the exporter logic every time new metric type is added? It's not scaling, especially when we need it on server side.

runtimes/runtimes/operational-telemetry/aws-metrics-exporter.ts
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            const operationalData = this.extractOperationalData(metrics)
+            this.sender
+                .sendOperationalTelemetryData(operationalData)
+                .then(() => {
+                    diag.info('Successfully exported operational metrics batch')
+                    resultCallback({ code: ExportResultCode.SUCCESS })
+                })
+                .catch(err => {
+                    diag.error('Failed to export operational metrics:', err)
```
volodkevych, line 41: Where will we see these errors?

runtimes/runtimes/operational-telemetry/telemetry-schemas/server-crash-schema.json
```diff
@@ -0,0 +1,20 @@
...
+        "ServerCrashEvent": {
+            "title": "ServerCrashEvent",
+            "allOf": [
+                { "$ref": "./base-event-schema.json" },
+                {
+                    "required": ["crashType"],
+                    "additionalProperties": false,
+                    "properties": {
+                        "crashType": {
+                            "type": "string"
```
volodkevych, line 13: I'm thinking it will be pretty difficult to troubleshoot a crash based on only the type of crash. I think we need a way to log related error stacktrace and error message. I do not think it should obligatory be part of metrics - it's rather logging details, but we should design this as well.

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        let attempt = 0
+        const maxRetries = 3
+        let delay = 1000 // Delay in milliseconds
+        let lastError: Error | null = null
+        let retry: boolean = true
+
+        for (; attempt <= maxRetries; attempt++) {
+            try {
+                const response = await axios.request({
```
ege0zcan, line 108: it seems like we are constructing a request twice, once inside `signRequest` function and once again inside this `axios.request` function. Is it possible to use the `signedRequest` object we already have and call `axios.request(signedRequest)` Or maybe we should modify signRequest function to return a request that is ready to be passed to axios.request

And one small nitpick: Can we use `axios.post` instead of `axios.request`?

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+                    method: 'POST',
+                    url: this.endpoint,
+                    data: body,
+                    headers: signedRequest.headers,
+                })
+
+                diag.debug(`Operational telemetry HTTP status: ${response.status}, message: ${response.statusText}`)
```
ege0zcan, line 115: maybe a stupid question but I want to make sure: do `diag.debug` logs get removed during bundling or do they become part of the bundle but logs don't show up unless we set an explicit env variable or something like that?

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+
+                return
+            } catch (error) {
+                lastError = error as Error
+
+                if (!retry || attempt === maxRetries) {
+                    break
+                }
+
+                // retries on network error from axios or 4xx error
```
ege0zcan, line 135: I don't understand what this comment means

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.test.ts
```diff
@@ -0,0 +1,82 @@
...
+            },
+        } as any)
+        diagStub = sinon.stub(diag, 'info')
+    })
+
+    afterEach(() => {
+        sinon.restore()
+    })
+
+    describe('sendOperationalTelemetryData', () => {
```
ege0zcan, line 42: Potential test cases for the future:
- sendOperationalTelemetryData retries with backoff on 429
- sendOperationalTelemetryData doesn't retry on 5xx and 4xx (except 429)

runtimes/runtimes/operational-telemetry/aws-metrics-exporter.ts
```diff
@@ -0,0 +1,124 @@
...
+                        | string
+                        | undefined,
+                },
+                clientId: this.telemetryService.getCustomAttributes()['clientInfo.clientId'] as string | undefined,
+            },
+            scopes: scopes,
+        }
+    }
+
+    private metricToOperationalEvent(metric: MetricData): ResourceUsageMetric {
```
ege0zcan, line 100: What happens in the future if we have more metric types? This function looks more like `metricToResourceUsageMetric` in its current form

runtimes/package.json
```diff
@@ -33,7 +33,20 @@
         "version": "npm run compile && git add -A ."
     },
     "dependencies": {
+        "@aws-crypto/sha256-js": "^5.2.0",
+        "@aws-sdk/client-cognito-identity": "^3.734.0",
+        "@aws-sdk/client-sts": "^3.734.0",
```
ege0zcan, line 38: nit: I don't think sts client is used right now

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    registerGaugeProvider(metricName: string, valueProvider: () => number, attributes?: Record<string, any>): void {
+        if (!this.initialized) {
+            diag.error('Operational telemetry not initialized')
+            return
+        }
+
+        const meter = opentelemetry.metrics.getMeter(this.SCOPE_NAME)
+        const gauge = meter.createObservableGauge(metricName.toString())
```
ege0zcan, line 45: isn't `metricName` already a string?

runtimes/runtimes/operational-telemetry/aws-metrics-exporter.ts
natdurlik: We initialize `diag` with `lspConnection.console ` in the `initialize` method. Output style of these logs will be the same as other usages of lspConnection.console(). For exemple:

`[Error - 6:57:41 PM] Failed to export operational metrics:
[Error - 6:57:41 PM] {"stack":"Error: PeriodicExportingMetricReader: metrics export failed (error undefined)\n    at PeriodicExportingMetricReader._doRun (/Users/natdur/flare/language-servers/node_modules/@opentelemetry/sdk-metrics/build/src/export/PeriodicExportingMetricReader.js:86:19)\n    at runNextTicks (node:internal/process/task_queues:60:5)\n    at listOnTimeout (node:internal/timers:545:9)\n    at process.processTimers (node:internal/timers:519:7)\n    at async PeriodicExportingMetricReader._runOnce (/Users/natdur/flare/language-servers/node_modules/@opentelemetry/sdk-metrics/build/src/export/PeriodicExportingMetricReader.js:54:13)","message":"PeriodicExportingMetricReader: metrics export failed (error undefined)","name":"Error"}`

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
natdurlik: yes, it was misleading

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+                    method: 'POST',
+                    url: this.endpoint,
+                    data: body,
+                    headers: signedRequest.headers,
+                })
+
+                diag.debug(`Operational telemetry HTTP status: ${response.status}, message: ${response.statusText}`)
+
+                if (response.status >= 400) {
+                    if (response.status < 500 && response.status != 429) {
```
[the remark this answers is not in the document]
natdurlik: refactored

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
natdurlik: `diag.debug` will work like `lspConnection.console.debug()`.

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+            throw new Error('No credentials received')
+        }
+
+        this.credentialsLastFetched = new Date()
+        this.credentials = credentialsResponse.Credentials
+
+        return this.credentials
+    }
+
+    private signRequest(url: URL, body: string, region: string, credentials: Credentials) {
```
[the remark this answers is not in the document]
natdurlik: makes sense, lets use `this`

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        const credentialsResponse = await cognitoIdentity.send(
+            new GetCredentialsForIdentityCommand({
+                IdentityId: identityResponse.IdentityId,
+            })
+        )
+
+        if (!credentialsResponse.Credentials) {
+            diag.error('Failed to refresh Cognito credentials')
```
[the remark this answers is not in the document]
natdurlik: same as `lspConnection.console.error()`

runtimes/runtimes/operational-telemetry/aws-cognito-gateway-sender.ts
```diff
@@ -0,0 +1,147 @@
...
+        const cognitoIdentity = new CognitoIdentityClient({ region })
+        const identityResponse = await cognitoIdentity.send(
+            new GetIdCommand({
+                IdentityPoolId: poolId,
+            })
+        )
+
+        const credentialsResponse = await cognitoIdentity.send(
+            new GetCredentialsForIdentityCommand({
+                IdentityId: identityResponse.IdentityId,
```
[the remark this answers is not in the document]
natdurlik: I assume it can throw an error and we will catch it in the exporter export method.

runtimes/runtimes/operational-telemetry/telemetry-schemas/telemetry-schema.json
```diff
@@ -0,0 +1,89 @@
...
+                    "type": "string"
+                }
+            }
+        },
+        "clientInfo": {
+            "type": "object",
+            "description": "Information about the client using the server",
+            "additionalProperties": false,
+            "properties": {
+                "name": {
```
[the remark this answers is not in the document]
natdurlik: as far as I understand client name is IDE/host platform and extension is extension. Fields are mapped from `ExtendedClientInfo` from our lsp protocol.

runtimes/runtimes/operational-telemetry/telemetry-schemas/server-crash-schema.json
natdurlik: We have a separate task for figuring out that part.

runtimes/runtimes/operational-telemetry/operational-event-validator.ts
```diff
@@ -0,0 +1,25 @@
+import Ajv from 'ajv'
+
+export class OperationalEventValidator {
+    private readonly ajv
+    private readonly validate
+
+    constructor() {
+        this.ajv = new Ajv()
+
+        this.ajv.addSchema(require('./telemetry-schemas/base-event-schema.json'), 'base-event-schema.json')
```
volodkevych, line 10: Can be more generic by iterating through all files in `telemetry-schemas` folder.

runtimes/runtimes/operational-telemetry/operational-event-validator.test.ts
```diff
@@ -0,0 +1,59 @@
...
+                name: 'ServerCrashEvent',
+                timestamp: 100000,
+                crashType: 'Error',
+            }
+
+            const result = validator.validateEvent(event)
+            assert.ok(result)
+        })
+
+        it('should return false when missing fields', () => {
```
volodkevych, line 49: What's the strategy for fields not in schema? Should we remove them or fail validation?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -0,0 +1,133 @@
...
+    private static instance: OperationalTelemetryService
+    private readonly RUNTIMES_SCOPE_NAME = 'language-server-runtimes'
+
+    static getInstance(
+        serviceName: string,
+        serviceVersion: string,
+        lspConsole: RemoteConsole,
+        poolId: string,
+        region: string,
+        endpoint: string
```
volodkevych, line 26: nit: there could be some config type introduced to avoid this many parameter

volodkevych (approved):

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -0,0 +1,80 @@
...
+        attributes?: Record<string, any>,
+        scopeName?: string
+    ): void
+    recordEvent(eventType: string, attributes?: Record<string, any>, scopeName?: string): void
+    getCustomAttributes(): Record<string, any>
+    updateCustomAttributes(key: string, value: any): void
+}
+
+export class OperationalTelemetryProvider {
+    private static telemetryInstance: OperationalTelemetry
```
volodkevych, line 14: why not allow to set it in constructor?

runtimes/runtimes/operational-telemetry/operational-event-validator.test.ts
natdurlik: I think we should fail validation, there is a special configuration in the json schema for that, namely `unevaluatedProperties: false`. Unfortunately it's not available yet in http://json-schema.org/draft-07/schema# that we use, and I couldn't get the `ajv` library work with the newer schema. We can improve it in the future.

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    static setTelemetryInstance(telemetryInstance: OperationalTelemetry): void {
+        OperationalTelemetryProvider.telemetryInstance = telemetryInstance
+    }
+
+    static getTelemetryForScope(scopeName: string): OperationalTelemetry {
+        if (!OperationalTelemetryProvider.telemetryInstance) {
+            return new NoopOperationalTelemetry()
```
volodkevych, line 22: Did we do it for webworker runtime? Or just because of `setTelemetryInstance`?

viktorsaws (approved):

viktorsaws (approved):

PR #316, opened 2025-02-03
feat: implement sdk runtime configurator to configure sdk clients at runtime

francescoopiccoli:
## Problem
We need to be able to set runtime/environment specific options for aws sdk clients at runtime.

## Solution
Configure sdk clients at runtime, by wrapping them into a sdk client configurator, which will use the specific sdk client constructor and options passed from the capability servers to initialize the clients at runtime. he sdk runtime configurator allows us to set additional runtime/environment-specific options (e.g. proxy). The sdk client configurator should be generic enough to work with any sdk clients, both v2 and v3.

## Important notes

This PR is connected to [this PR](https://github.com/aws/language-servers/pull/767) in language-servers.

I locally tested the changes configuring proxy at runtime, it works fine.

Setting a generic type for sdk v3 clients (so that the solution would work with (and exclusively) any kind of aws sdk v3 client) has proven particularly tricky. The best solution I could find is the one implemented here, with the only downside of needing typecasting in the capability servers (see [here](https://github.com/aws/language-servers/pull/767/commits/8cccbd4ae92d7787b1037c68c822fc857e6d7d67#diff-7122529b7d94d974b59dff123163f016147d4f2ef8606d8f35781bcb811616e1R58-R65)).

Another solution to avoid type casting would be removing the `extends SDKClientV3` so:
`v3: <T, P>(Ctor: SDKClientConstructorV3<T, P>, current_config: P) => T` which has the downside of giving more freedom and less constraints on the constructor type, hence letting any possible object to be passed as input. 

I also tried to play around with smithy client, I found another working solution:
```
import { Client as __Client } from "@smithy/smithy-client";

export type SDKClientV3<ClientInput extends object, ClientOutput extends MetadataBearer> = __Client<HttpHandlerOptions, ClientInput, ClientOutput, SmithyResolvedConfiguration<HttpHandlerOptions>>
v3: <T extends SDKClientV3<any, any>, P>(Ctor: SDKClientConstructorV3<T, P>, current_config: P) => T
```
this solution has the downside of introducing a dependency to `smithy/smithy-client`, it does not require typeacasting for the clients we currently use, but it does require it for some other clients, so it is not much better then the currently implemented solution. 

Let's discuss this offline if needed.

## License


runtimes/runtimes/util/proxyUtil.ts
francescoopiccoli, line 1: v2 proxy config setup taken from: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/utils.ts#L134C14-L134C29
v3 proxy config setup taken from: https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/proxy-server.ts#L48

runtimes/server-interface/sdk-decorator.ts
```diff
@@ -0,0 +1,58 @@
+import { Service } from 'aws-sdk'
+import { ServiceConfigurationOptions } from 'aws-sdk/lib/service'
+import { Client, MetadataBearer, HttpHandlerOptions } from '@aws-sdk/types'
+import { SmithyResolvedConfiguration } from '@aws-sdk/smithy-client'
+
+// aws sdk v2 clients constructor type
+export type SDKClientConstructorV2<T, P extends ServiceConfigurationOptions> = new (config?: P) => T
+// aws sdk v3 clients constructor type
+export type SDKClientConstructorV3<T, P> = new (...[configuration]: [P] | []) => T
```
imykhai, line 9: Aren't the types equivalent? 
`new (config?: P) => T` 
and 
`new (...[configuration]: [P] | []) => T` 

Are there scenarios when they are not the same?

runtimes/server-interface/sdk-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// a generic type for all aws sdk v3 clients.
+export type SDKClientV3 = Client<object, MetadataBearer, SmithyResolvedConfiguration<HttpHandlerOptions>>
```
imykhai, line 12: Why are we specializing  `SmithyResolvedConfiguration` with the `HttpHandlerOptions`? Does it allow other configuration to be passed to the client?

runtimes/server-interface/sdk-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+     * @example
+     * // Example usage:
+     * const mySdkV3Client = sdkRuntimeConfigurator.v3(
+     *   MySdkV3Client as any,
+     *   {
+     *       region: 'example_region',
+     *       endpoint: 'example_endpoint',
+     *   }) as MySdkV3Client;
+     */
+    v3: <T extends SDKClientV3, P>(Ctor: SDKClientConstructorV3<T, P>, current_config: P) => T
```
imykhai, line 57: Can we make `v3` the *default* runtime configurator? Maybe by making the `configurator` a function itself that accepts the v3 sdk clients, while `v2` will still be a method on the `configurator` object.  I'm basing this on the idea that the v2 will get deprecated soon and expecting the capability developers to call v3 will be redundant.

runtimes/server-interface/sdk-decorator.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// aws sdk v2 clients constructor type
+export type SDKClientConstructorV2<T, P extends ServiceConfigurationOptions> = new (config?: P) => T
+// aws sdk v3 clients constructor type
+export type SDKClientConstructorV3<T, P> = new (...[configuration]: [P] | []) => T
+
+// a generic type for all aws sdk v3 clients.
+export type SDKClientV3 = Client<object, MetadataBearer, SmithyResolvedConfiguration<HttpHandlerOptions>>
+
+export type SDKRuntimeConfigurator = {
```
imykhai, line 14: I'm thinking if we should omit the `Runtime` part. This can be misinterpreted as if we're talking about some `SDK runtime` which isn't the case.

runtimes/server-interface/sdk-decorator.ts
francescoopiccoli: In practice they achieve the same thing, but this is needed:

V3 example constructor:  https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-s3/src/S3Client.ts#L828
V2 example constructor: https://github.com/aws/aws-sdk-js/blob/master/clients/s3.d.ts#L18

If you try to use `new (config?: P) => T` on the v3 constructor, you get an error:
```
Types of construct signatures are incompatible.
    Type 'new (...[configuration]: [] | [sdkV3ClientConfig]) => sdkV3Client' is not assignable to type 'new (config?: {...
```

So the constructor while very similar are needed for different constructor parameter handling for v2 and v3 clients

runtimes/server-interface/sdk-decorator.ts
francescoopiccoli: `HttpHandlerOptions` is the default for all v3. (see [here](https://github.com/aws/aws-sdk-js-v3/blob/2b3d1df4f1343b09d8a7200c0d62cf87de35e8ed/UPGRADING.md?plain=1#L54-L58)).

I checked a few v3 clients, they all use that ([example](https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-s3/src/S3Client.ts#L793)).

The other option that would not assume the default `HttpHandlerOptions` would be to having the capability servers passing down the type parameter for SmithyResolvedConfiguration, but I prefer this solution for the same reason that we discussed yesterday abou avoiding type casting

runtimes/server-interface/sdk-decorator.ts
francescoopiccoli: In my view, adding `Runtime` made it clear that we are adding/initializing the sdk client with some additional configuration at runtime. 
Removing the `Runtime` part would raise the question to who looks at the code "Why is there a configurator function taking constructor and options when i can initialize the client here with same constructor and options?"
Let me know if this makes sense to you, I have no strong opinion, in any case i can add some extra documentation to make the goal and scope of the configurator clear, irrespective of which naming we choose

runtimes/server-interface/sdk-decorator.ts
francescoopiccoli: Good point. So it would be something like this?
```
export type SDKRuntimeConfigurator = {
    v2: <T extends Service, P extends ServiceConfigurationOptions>(Ctor: SDKClientConstructorV2<T, P>, current_config: P) => T;
} & (<T, P>(Ctor: SDKClientConstructorV3<T, P>, current_config: P) => T);

// V3 usage (default)
const v3Client = sdkRuntimeConfigurator(MySdkV3Client, {
    region: 'example_region',
    endpoint: 'example_endpoint'
});

// V2 usage (still supported)
const v2Client = sdkRuntimeConfigurator.v2(MySdkV2Client, {
    region: 'example_region',
    endpoint: 'example_endpoint'
});
```
Somewhat i don't like the lack of symmetry between v2 and v3. But you have a point regarding making v3 default, despite v2 end of support is still months away. I am good with the new proposed solution if you think it's better!

francescoopiccoli: Alternatively if we go full generic also v2, we could have a single function:
```
export type SDKRuntimeConfigurator = <T, P>(
    Ctor: SDKClientConstructorV2<T, P> | SDKClientConstructorV3<T, P>, 
    current_config: P, 
    isV2?: boolean
) => T;
```
but i would be against avoiding types also for v2

runtimes/server-interface/sdk-decorator.ts
imykhai: we definitely need to document why the capability developers should abstain from instantiating the SDK Clients themselves, but instead to pass it on to the `SDKRuntimeConfigurator`. 
Also, previously we thought that the configurator will only apply some additional settings to an already created SDK object, but now the configurator also instantiates the object. Should we therefore consider `SDKInitializator` or similar?

runtimes/server-interface/sdk-decorator.ts
imykhai: I don't have strong preferences either. 
If you agree let's go with the v2 being default as in your first code snippet above.

runtimes/server-interface/sdk-decorator.ts
imykhai, line 1: let's keep the file name consistent with the latest naming "sdk-intializator"

imykhai (approved):

PR #323, opened 2025-02-10
feat: added agent protocol

volodkevych:
## License

### DIFF
--- a/runtimes/protocol/agent.ts
+++ b/runtimes/protocol/agent.ts
@@ -4,13 +4,13 @@ import {
     ManageTaskResult,
     ProtocolNotificationType,
     ProtocolRequestType,
-    TASK_STATUS_UPDATE_NOTIFICATION_METHOD,
-    TaskStatusUpdateParams,
+    TASK_STATE_UPDATE_NOTIFICATION_METHOD,
+    TaskStateUpdateParams,
 } from './lsp'
 
 export const manageTaskRequestType = new ProtocolRequestType<ManageTaskParams, ManageTaskResult, never, void, void>(
     MANAGE_TASK_REQUEST_METHOD
 )
-export const sendTaskStatusUpdateNotificationType = new ProtocolNotificationType<TaskStatusUpdateParams, void>(
-    TASK_STATUS_UPDATE_NOTIFICATION_METHOD
+export const sendTaskStateUpdateNotificationType = new ProtocolNotificationType<TaskStateUpdateParams, void>(
+    TASK_STATE_UPDATE_NOTIFICATION_METHOD
 )
--- a/runtimes/runtimes/webworker.ts
+++ b/runtimes/runtimes/webworker.ts
@@ -33,7 +33,7 @@ import {
     openFileDiffNotificationType,
     selectWorkspaceItemRequestType,
     manageTaskRequestType,
-    sendTaskStatusUpdateNotificationType,
+    sendTaskStateUpdateNotificationType,
     openTabRequestType,
 } from '../protocol'
 import { BrowserMessageReader, BrowserMessageWriter, createConnection } from 'vscode-languageserver/browser'
@@ -213,8 +213,8 @@ export const webworker = (props: RuntimeProps) => {
 
         const agent: Agent = {
             onManageTask: handler => lspConnection.onRequest(manageTaskRequestType.method, handler),
-            sendTaskStatusUpdate: params =>
-                lspConnection.sendNotification(sendTaskStatusUpdateNotificationType.method, params),
+            sendTaskStateUpdate: params =>
+                lspConnection.sendNotification(sendTaskStateUpdateNotificationType.method, params),
         }
 
         const sdkInitializator: SDKInitializator = Object.assign(
--- a/runtimes/server-interface/agent.ts
+++ b/runtimes/server-interface/agent.ts
@@ -1,10 +1,4 @@
-import {
-    ManageTaskParams,
-    ManageTaskResult,
-    NotificationHandler,
-    RequestHandler,
-    TaskStatusUpdateParams,
-} from '../protocol'
+import { ManageTaskParams, ManageTaskResult, RequestHandler, TaskStateUpdateParams } from '../protocol'
 
 export type Agent = {
     // Requests
@@ -12,5 +6,5 @@ export type Agent = {
         handler: RequestHandler<ManageTaskParams, ManageTaskResult | undefined | null, ManageTaskResult>
     ) => void
     // Notifications
-    sendTaskStatusUpdate: (params: TaskStatusUpdateParams) => void
+    sendTaskStateUpdate: (params: TaskStateUpdateParams) => void
 }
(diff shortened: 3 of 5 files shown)


types/workspace.ts
```diff
@@ -0,0 +1,23 @@
+import { URI } from './lsp'
+
+export const SELECT_WORKSPACE_ITEM_REQUEST_METHOD = 'aws/selectWorkspaceItem'
+export const OPEN_FILE_DIFF_NOTIFICATION_METHOD = 'aws/openFileDiff'
+
+export interface SelectWorkspaceItemParams {
+    canSelectFolders: boolean
```
jpinkney-aws, line 7: Have you checked with JB/Eclipse/VS to see if these options are more or less supported on their sides?

types/chat.ts
```diff
@@ -186,8 +199,21 @@ export interface SourceLinkClickParams extends InfoLinkClickParams {
...
+    placeholderText?: string
+    canAcceptInput?: boolean
+    cardAction?: 'Add' | 'UpdateLast' | 'UpdateByMessageId'
+}
+
+export interface FileClickParams extends Partial<TaskParams> {
+    tabId: string
+    filePath: string
+    isDeleted: boolean
+    action?: FileAction
```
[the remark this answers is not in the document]
volodkevych: It's needed for 'accept-change' action, which needs to be handled on server side to trigger workspace edit.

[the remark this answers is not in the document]
dogusata: what about sub-actions?

runtimes/protocol/agent.ts
```diff
@@ -0,0 +1,16 @@
...
+    MANAGE_TASK_REQUEST_METHOD,
+    ManageTaskParams,
+    ManageTaskResult,
+    ProtocolNotificationType,
+    ProtocolRequestType,
+    TASK_STATUS_UPDATE_NOTIFICATION_METHOD,
+    TaskStatusUpdateParams,
+} from './lsp'
+
+export const manageTaskRequestType = new ProtocolRequestType<ManageTaskParams, ManageTaskResult, never, void, void>(
```
kmile, line 11: `ManageTask` is a bit too generic. Even when reading through `ManageTaskParams` it's not super clear what kind of management operations we'll get. If we only have `Start` and `Stop`, I'd prefer to model those explicitly in 2 requests. If we have any kinds of future states a task could transition to, then this should be something more like `Call` with an `Action` so those transitions can be modeled.

runtimes/protocol/workspace.ts
```diff
@@ -0,0 +1,20 @@
...
+    SelectWorkspaceItemResult,
+} from './lsp'
+
+export const selectWorkspaceItemRequestType = new ProtocolRequestType<
+    SelectWorkspaceItemParams,
+    SelectWorkspaceItemResult,
+    never,
+    void,
+    void
+>(SELECT_WORKSPACE_ITEM_REQUEST_METHOD)
```
kmile, line 17: If this is using a request type, params, and result from `./lsp`, should it be defined in `workspace`? Is this responsibility split between these two namespaces and if so, can we make the distinction more clear why this belongs here and all the data types don't?

kmile:
A readme.md at the top of the `/agent` folder could help to show how agents work, what responsibilities they have, and the key requests/responses/notifications they support. I know this is an early draft but it could help with the review as well to understand the bigger picture.

runtimes/runtimes/webworker.ts
```diff
@@ -57,7 +64,7 @@ import {
     updateProfileRequestType,
 } from '../protocol/identity-management'
 import { IdentityManagement } from '../server-interface/identity-management'
-import { Encoding, WebBase64Encoding } from './encoding'
+import { WebBase64Encoding } from './encoding'
```
kmile, line 67: Unused?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,16 @@
...
+    TaskStatusUpdateParams,
+} from '../protocol'
+
+export type Agent = {
+    // Requests
+    onManageTask: (
+        handler: RequestHandler<ManageTaskParams, ManageTaskResult | undefined | null, ManageTaskResult>
+    ) => void
+    // Notifications
+    sendTaskStatusUpdate: (handler: NotificationHandler<TaskStatusUpdateParams>) => void
```
kmile, line 15: This isn't a `handler`, is it? Is this just a `params => void` method call?

types/agent.ts
```diff
@@ -0,0 +1,43 @@
...
+    detail?: string
+}
+export interface TaskFileList {
+    filePaths?: string[]
+    deletedFiles?: string[]
+}
+export interface TaskStep {
+    step: string
+    detail?: string
+    fileList?: TaskFileList
```
kmile, line 32: Why is this explicitly a part of `TaskStep`? I know this is optional, but if we model every effect a task can have as optional parameter I can see this type growing very rapidly over time.

types/agent.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+export interface TaskAction {
+    type: TaskActionType
+}
+export interface TaskStatusUpdate {
+    overview: TaskOverview
+    status: TaskStatus
+    step?: TaskStep
```
kmile, line 40: Should this be a list of past, current, and upcoming steps? To see what was already done and what still remains? I know not every task knows all steps up front, but a lot of them do.

runtimes/protocol/agent.ts
volodkevych: `TaskStatusUpdate` provides a list of actions, meaning that it's next available actions for the current task. Is it what you meant by "modeled"?

runtimes/protocol/chat.ts
```diff
@@ -32,6 +32,10 @@ import {
     EncryptedChatParams,
     EncryptedQuickActionParams,
     QuickActionResult,
+    TASK_UPDATE_NOTIFICATION_METHOD,
+    SendUpdateParams,
```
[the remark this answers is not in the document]
volodkevych: I removed "Task" from there, because there's nothing Task-specific about it. It can theoretically be used to update chat state for any async flow.

runtimes/runtimes/webworker.ts
volodkevych: yes

runtimes/protocol/workspace.ts
volodkevych: All the protocol types are re-exported from `./lsp`. May be worth revisiting this decision to avoid confusion.

types/agent.ts
volodkevych: I've seen agents using files to report results is kind of a common pattern. Will try to clarity this with agent teams.

types/agent.ts
volodkevych: I'll change this to array. But I'm not sure about reporting all the steps every time vs reporting delta. Reporting all the steps every time could be excessive in my opinion.

justinmk3:
is this replaced by https://github.com/aws/language-server-runtimes/pull/368 ?

volodkevych:
Yes, replaced by https://github.com/aws/language-server-runtimes/pull/368

PR #325, opened 2025-02-12
feat(runtimes): extract reusable runtime logic into the `base-runtime.ts`

imykhai:
## Problem
Neither `webworker` runtime nor `standalone` runtime allow passing a different transport mechanism, `Webworker` use `postMessage` (BrowserMessageReader and BrowserMessageWriter) while `standalone` uses `stdin` and `stdout` as the transport for JSON-RPC. 

## Solution
In this PR I'm introducing `base-runtime` that can accept custom `reader` and `writer`. `webworker` runtime now just passes the `postMessage`-based transport into the base runtime. 
The next step will be to also align the base-runtime with the `standalone` and then use it as a basis for that runtime too. 

`base-runtime.ts` almost fully consists of the code that was previously in the `webworker.ts` with the only difference of accepting `reader` and `writer` as the parameters. 

## License


runtimes/runtimes/base-runtime.ts
```diff
@@ -0,0 +1,236 @@
+import { TextDocuments } from 'vscode-languageserver'
```
natdurlik, line 1: is the base-runtime only for web runtimes or also for the standalone?

runtimes/runtimes/base-runtime.ts
```diff
@@ -0,0 +1,236 @@
...
+} from '../protocol/identity-management'
+import { IdentityManagement } from '../server-interface/identity-management'
+import { Encoding, WebBase64Encoding } from './encoding'
+import { LoggingServer } from './lsp/router/loggingServer'
+import { Service } from 'aws-sdk'
+import { ServiceConfigurationOptions } from 'aws-sdk/lib/service'
+
+declare const self: WindowOrWorkerGlobalScope
+
+export const baseRuntime = (connections: { reader: MessageReader; writer: MessageWriter }) => (props: RuntimeProps) => {
```
francescoopiccoli, line 69: nit: should we add a couple of lines above to document?

francescoopiccoli (approved):

runtimes/runtimes/base-runtime.ts
imykhai: In this PR I only use it for the web runtimes. But we should followup with standalone runtime switching to it too.

runtimes/runtimes/base-runtime.ts
imykhai: added

PR #330, opened 2025-02-17
feat: add opt-out mechanism to operational telemetry

natdurlik:
## Problem
We need to respect customer decision and provide a way for them to opt-out of the operational telemetry.

## Solution
- Extended the telemetry interface with a method to update telemetry opt out status, which determines if metrics are collected or not.
- Added another LSP server to the router that updates telemetry opt-out status upon initialization and when workspace configuration variables change.
- The OpenTelemetry SDK is shut down when a user turns off telemetry and started again when a user enables telemetry.
- Enabling or disabling telemetry does not require an IDE/extension restart.

## License

### DIFF
--- a/runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
+++ b/runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
@@ -1,24 +1,17 @@
 import { ReadableSpan, SpanExporter } from '@opentelemetry/sdk-trace-base'
 import { ExportResult, ExportResultCode } from '@opentelemetry/core'
 import { AwsCognitoApiGatewaySender } from './aws-cognito-gateway-sender'
-import { OperationalTelemetry, TelemetryStatus } from './operational-telemetry'
 import { diag } from '@opentelemetry/api'
 import { OperationalEvent, OperationalTelemetrySchema } from './types/generated/telemetry'
 import { OperationalEventValidator } from './operational-event-validator'
 
 export class AwsSpanExporter implements SpanExporter {
-    private readonly telemetryService: OperationalTelemetry
     private readonly sender: AwsCognitoApiGatewaySender
     private readonly eventValidator: OperationalEventValidator
 
     private isShutdown = false
 
-    constructor(
-        telemetryService: OperationalTelemetry,
-        sender: AwsCognitoApiGatewaySender,
-        eventValidator: OperationalEventValidator
-    ) {
-        this.telemetryService = telemetryService
+    constructor(sender: AwsCognitoApiGatewaySender, eventValidator: OperationalEventValidator) {
         this.sender = sender
         this.eventValidator = eventValidator
     }
@@ -27,17 +20,6 @@ export class AwsSpanExporter implements SpanExporter {
         if (this.isShutdown) {
             diag.warn('Export attempted on shutdown exporter')
             resultCallback({ code: ExportResultCode.FAILED })
-            return
-        }
-        if (this.telemetryService.getTelemetryStatus() === TelemetryStatus.Pending) {
-            // add to queue
-            resultCallback({ code: ExportResultCode.SUCCESS })
-            return
-        }
-        if (this.telemetryService.getTelemetryStatus() === TelemetryStatus.Disabled) {
-            // clean queue
-            diag.warn('Telemetry is disabled, dropping metrics')
-            resultCallback({ code: ExportResultCode.SUCCESS })
             return
         }
         if (spans.length === 0) {
@@ -103,16 +85,13 @@ export class AwsSpanExporter implements SpanExporter {
                 version: spans[0].resource.attributes['service.version'] as string | undefined,
             },
             clientInfo: {
-                name: this.telemetryService.getCustomAttributes()['clientInfo.name'] as string | undefined,
+                name: spans[0].resource.attributes['clientInfo.name'] as string | undefined,
+                version: spans[0].resource.attributes['clientInfo.version'] as string | undefined,
                 extension: {
-                    name: this.telemetryService.getCustomAttributes()['clientInfo.extension.name'] as
-                        | string
-                        | undefined,
-                    version: this.telemetryService.getCustomAttributes()['clientInfo.extension.version'] as
-                        | string
-                        | undefined,
+                    name: spans[0].resource.attributes['clientInfo.extension.name'] as string | undefined,
+                    version: spans[0].resource.attributes['clientInfo.extension.version'] as string | undefined,
                 },
-                clientId: this.telemetryService.getCustomAttributes()['clientInfo.clientId'] as string | undefined,
+                clientId: spans[0].resource.attributes['clientInfo.clientId'] as string | undefined,
             },
             scopes: scopes,
         }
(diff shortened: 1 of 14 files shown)


runtimes/runtimes/util/telemetryLspServer.ts
```diff
@@ -0,0 +1,37 @@
...
+        }
+
+        OperationalTelemetryProvider.getTelemetryForScope('').updateTelemetryStatus(status)
+
+        return {
+            capabilities: {},
+        }
+    })
+    lspServer.setDidChangeConfigurationHandler(async params => {
+        const optOut = (await lspConnection.workspace.getConfiguration({
```
volodkevych, line 24: Will client have to pass `aws.q.optOutTelemetry` section on every configuration change?

runtimes/runtimes/standalone.ts
```diff
@@ -64,6 +64,7 @@ import { Encoding } from './encoding'
 import { LoggingServer } from './lsp/router/loggingServer'
 import { Service } from 'aws-sdk'
 import { ServiceConfigurationOptions } from 'aws-sdk/lib/service'
+import { getTelmetryLspServer } from './util/telemetryLspServer'
```
volodkevych, line 67: typo

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
```diff
@@ -29,6 +29,17 @@ export class AwsSpanExporter implements SpanExporter {
             resultCallback({ code: ExportResultCode.FAILED })
             return
         }
+        if (this.telemetryService.getTelemetryStatus() === TelemetryStatus.Pending) {
+            // add to queue
+            resultCallback({ code: ExportResultCode.SUCCESS })
+            return
+        }
+        if (this.telemetryService.getTelemetryStatus() === TelemetryStatus.Disabled) {
```
volodkevych, line 37: Why is it needed - isn't it enough to shutdown `NodeSDK`?

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
```diff
@@ -29,6 +29,17 @@ export class AwsSpanExporter implements SpanExporter {
             resultCallback({ code: ExportResultCode.FAILED })
             return
         }
+        if (this.telemetryService.getTelemetryStatus() === TelemetryStatus.Pending) {
```
volodkevych, line 32: Really a corner case, but should we make sure we do not have memory leak because of this, if logic with status change is not working?

runtimes/runtimes/util/telemetryLspServer.ts
```diff
@@ -0,0 +1,37 @@
+import { Connection, InitializeParams, InitializeResult } from 'vscode-languageserver/node'
+import { Encoding } from '../encoding'
+import { Logging } from '../../server-interface/logging'
+import { LspServer } from '../lsp/router/lspServer'
+import { OperationalTelemetryProvider, TelemetryStatus } from '../operational-telemetry/operational-telemetry'
+
+export function getTelmetryLspServer(lspConnection: Connection, encoding: Encoding, logging: Logging): LspServer {
+    const lspServer = new LspServer(lspConnection, encoding, logging)
+    lspServer.setInitializeHandler(async (params: InitializeParams): Promise<InitializeResult> => {
+        const optOut: boolean = params.initializationOptions?.telemetryOptOut ?? false
```
volodkevych, line 10: Should we differentiate between `undefined` and `false`. Imagine clients who get new version of LSP server and do not set `telemetryOptOut` yet. I do not think we want to increment major version because of this.

volodkevych:
Tests would be useful

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
natdurlik: I've considered the corner cases in this scenario. When `sdk.shutdown()` is called, it triggers `metricReader.shutdown()`, which in turn initiates a flush of collected metrics (https://github.com/open-telemetry/opentelemetry-js/blob/e9d3c71918635d490b6a9ac9f8259265b38394d0/packages/sdk-metrics/src/export/PeriodicExportingMetricReader.ts#L164)). 
This if statement ensures that even during this shutdown process, these final metrics won't be sent if telemetry is disabled.

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
natdurlik: Would a queue size limit be an effective solution here?

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -6,10 +6,18 @@ export interface OperationalTelemetry {
         scopeName?: string
     ): void
     recordEvent(eventType: string, attributes?: Record<string, any>, scopeName?: string): void
+    updateTelemetryStatus(status: TelemetryStatus): void
+    getTelemetryStatus(): TelemetryStatus
     getCustomAttributes(): Record<string, any>
     updateCustomAttributes(key: string, value: any): void
 }
 
+export enum TelemetryStatus {
```
ege0zcan, line 15: nit: worth investigating the difference between using enums and typescript types here for our use case

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -6,10 +6,18 @@ export interface OperationalTelemetry {
...
+    Disabled,
+    Enabled,
+    Pending,
```
ege0zcan, line 18: One approach that we discussed offline is the option to get rid of the pending state and not collecting any metrics until the server initialization has happened. 

That approach would simplify the logic and we would only lose metrics that happen before server initialization which I would argue are not very useful for our purposes anyway

natdurlik: Changed to this approach. We ca always add collection before the initialize later if we find it useful.

runtimes/runtimes/util/telemetryLspServer.ts
natdurlik: right, changed it to opted-out if not provided.

ege0zcan (approved):

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -11,57 +11,99 @@ import { AwsCognitoApiGatewaySender } from './aws-cognito-gateway-sender'
 import { AwsSpanExporter } from './aws-spans-exporter'
 import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base'
 import { OperationalEventValidator } from './operational-event-validator'
+import { ExtendedClientInfo } from '../../server-interface'
+
+type OperationalTelmetryConfig = {
```
ege0zcan, line 16: nit: typo here

ege0zcan (approved):

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -11,57 +11,99 @@ import { AwsCognitoApiGatewaySender } from './aws-cognito-gateway-sender'
 import { AwsSpanExporter } from './aws-spans-exporter'
 import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base'
 import { OperationalEventValidator } from './operational-event-validator'
+import { ExtendedClientInfo } from '../../server-interface'
+
+type OperationalTelemetryConfig = {
+    serviceName: string
+    serviceVersion?: string
+    extendedClientInfo?: ExtendedClientInfo
+    lspConsole: RemoteConsole
```
volodkevych, line 20: nit: it should not be part of config

volodkevych (approved):

PR #335, opened 2025-02-19
feat(runtimes): extended network proxy support

viktorsaws:
## Problem
Standalone runtime runs as a NodeJS process on customers environment. There are customers with different network configurations, which impacts connectivity of nodejs process to internet:
* Explicit use of HTTP(s) proxy 
* Transparent proxy configured on Network level (common for corporate environments)

Runtime need configure instances of SDKs to work properly behind proxies for both cases.

## Solution

1. [x] Update proxy configurations of AWS SDKs to add support for both network configuration use cases to add support for transparent proxies.
2. [x] Add support for HTTP proxy values set in env variables to include `HTTP_PROXY` and `http_proxy` values.
3. [x] Read certificates from OS trust store, with support of MacOS, Windows and Linux OS-es. Maintain reading certificates passed in `NODE_EXTRA_CA_CERTS` or `AWS_CA_BUNDLE` env variables. Configure AWS SDKs with http agent, using read valid certificates.
5. [ ] Update documentation for proxy support and distribution of extra files for standalone bundles.

### Note of Windows certificate reader
Using [win-ca](https://github.com/ukoloff/win-ca) npm package to read certificates. All solutions I found for reading certificates on NodeJS on Windows use [Node-API](https://nodejs.org/api/n-api.html), which require distributing compiled add-on for Node in `.node` format. `win-ca` provides 2 add-ons, which need to be shipped with every standalone distribution bundle to work properly on Windows.

### Explicit proxy

In this solution, runtimes will read and cache certificates at NodeJS process startup.
At startup, server checks is proxy URL is set in one of next variables, with first found value having higher priority and used for proxy url: `HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy`. Proxy's certificates must be installed in Operation System certificates trust store or passed in `NODE_EXTRA_CA_CERTS` or `AWS_CA_BUNDLE` environmental variables.

*Note: if proxy url uses SSL and starts with `https://`, setting path to certificates in `NODE_EXTRA_CA_CERTS` is required before starting runtime process. Using  `NODE_EXTRA_CA_CERTS` is standard way to pass extra certificates in Nodejs.*

### Transparent network proxy

If you have transparent proxy configured in your network, it's certificates must be installed in the OS certificates trust store, or passed in `NODE_EXTRA_CA_CERTS` or `AWS_CA_BUNDLE` enviromental variables. Typically this setup exists in corporate environment, and certificates must be preinstalled on the machine. If not, customer needs to obtain certificates from Network administrator.

To support transparent proxies, runtime will read certificates from OS trust store, `NODE_EXTRA_CA_CERTS` and `AWS_CA_BUNDLE` env variables at startup.

## Test cases

More testing will be done on MacOS, Windows and Linux platform to ensure solution is working. Test matrix below is for reference.
Fully tested using https://mitmproxy.org/ on MacOS and Windows.

- [x] Transparent proxy - HTTP(s) proxy is configured by network administrator, and is not known to end user. Proxy certificates must be available to make HTTP calls.
  - [x]  No certificates - fail to connect
  - [x]  Certificate is installed in OS Trust Store
  - [x]  Path to Certificate is set in AWS_CA_BUNDLE env variable
  - [x]  Path to Certificate is set in NODE_EXTRA_CA_CERTS env variable
- [x] Explicit proxy set in env variables - HTTP
  - [x]  No certificates - fail to connect
  - [x]  Certificate is installed in OS Trust Store
  - [x]  Path to Certificate is set in AWS_CA_BUNDLE env variable
  - [x]  Path to Certificate is set in NODE_EXTRA_CA_CERTS env variable
- [x] Explicit proxy set in env variables - HTTPS
  - [x]  No certificates - fail to connect
  - [ ] Does not work: Certificate is installed in OS Trust Store 
  - [ ] Does not work: Path to Certificate is set in AWS_CA_BUNDLE env variable
  - [x] Path to Certificate is set in NODE_EXTRA_CA_CERTS env variable

Current solution has limited support for secure proxy, starting with `https://`. To make it work, it is required to set `NODE_EXTRA_CA_CERTS` to point to correct `.pem` certificate file, issued by proxy server. This is standard way of NodeJS to configure process for this use case. I was not able to configure proxy agent to support `https` proxy in code, so I'd add only partial support for now.

## License


runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,127 @@
...
+ * - AWS_CA_BUNDLE environment variable (if set)
+ * - NODE_EXTRA_CA_CERTS environment variable (if set)
+ *
+ * @returns {string[]} Array of certificate strings in PEM format
+ */
+function getCertificates(): string[] {
+    const certificates: string[] = []
+
+    const readCertificateFile = (path: string): void => {
+        try {
```
volodkevych, line 44: nit: could be useful to log if read cert based on env var happened or not

runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,127 @@
...
+function createSecureAgent(): HttpsAgent | HttpsProxyAgent {
+    const certs = getCertificates()
+    const agentOptions = {
+        ca: certs,
+        rejectUnauthorized: true, // Always validate certificates
+    }
+
+    const proxyUrl = getProxyUrl()
+
+    if (proxyUrl) {
```
volodkevych, line 84: Again, would be useful to log if we use proxyUrl or not.

runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,127 @@
...
+/**
+ * Creates an HTTPS agent configured with certificates and proxy settings if applicable.
+ * The agent is configured to always validate certificates for secure communication.
+ *
+ * @returns {HttpsAgent | HttpsProxyAgent} Configured HTTPS agent for making secure requests
+ */
+function createSecureAgent(): HttpsAgent | HttpsProxyAgent {
+    const certs = getCertificates()
+    const agentOptions = {
+        ca: certs,
```
volodkevych, line 78: If any certificate in this chain will be invalid or expired, will it skip it or fail?

viktorsaws: I don't' think the package I'm using is validating certificates, I'll check

runtimes/runtimes/util/standalone/proxyUtil.ts
viktorsaws: yeah, more logging is coming

justinmk3:
Interesting this could be useful for: 

- https://github.com/aws/aws-toolkit-vscode/issues/185

runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    readSystemCertificates(): string[] {
+        const platform = process.platform
+        let certs: string[] = []
+
+        if (platform === 'darwin') {
+            certs = readMacosCertificates()
+        } else if (platform === 'linux') {
+            certs = readLinuxCertificates()
+        } else if (platform === 'win32') {
```
volodkevych, line 86: Should we list what envs we support in readme?

runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,192 @@
...
+    private isValidProxyUrl(url: string): boolean {
+        try {
+            new URL(url)
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    private removeExpiredCertificates(certs: string[]) {
```
volodkevych, line 173: Is it really required? Won't it be handled by proxy agent?

runtimes/runtimes/util/standalone/proxyUtil.ts
```diff
@@ -0,0 +1,192 @@
...
+            rejectUnauthorized: true, // Always validate certificates
+        }
+
+        const proxyUrl = this.getProxyUrl()
+
+        // Proxy agent for explicit proxy
+        if (proxyUrl) {
+            console.log(`Using HTTP proxy at ${proxyUrl}`)
+
+            return new HttpsProxyAgent({
```
volodkevych, line 153: Have you checked `secureProxy` setting? I think we need to set it to true.

volodkevych (approved):

runtimes/runtimes/util/standalone/proxyUtil.ts
viktorsaws: yes, good idea, I'll update documentation I created

viktorsaws:
> Interesting this could be useful for:
> 
> * [Proxy Support aws-toolkit-vscode#185](https://github.com/aws/aws-toolkit-vscode/issues/185)

@justinmk3 this is common problem with NodeJS processes not picking up certificates configures locally or not respecting local proxy network configurations out of the box. Workaround for extensions in VSCode is to use one of those HTTPs certificates you found. But that solves problem for extension making HTTP requests. 

If there's some functionality that spawns separate nodejs processes (like flare language servers), those processes need to be configured with proxy config. Easiest solution for MITM proxies in customer environment is to set system-wide `NODE_EXTRA_CA_CERTS`, pointing to proxy's certificate.

PR #332, opened 2025-02-17
chore: add initializationOption data to error log and metrics

rahmaniaam:
## Problem
Previously we added logging and emitted a metric when initialization failed due to missing fields in `initializationOptions`, it was however not verbose enough to be able to pinpoint the exact issue
## Solution
Emit and log the exact full `initializationOptions` data

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #310, opened 2025-01-28
fix: explicit check for initializationOptions and add optional chaining

suprajaven:
## Problem
Check related PRs https://github.com/aws/language-server-runtimes/pull/306 and https://github.com/aws/language-server-runtimes/pull/307
Prevent errors due to aws object being undefined and improve error handling in initialize method

## Solution
- Updating to use optional chaining  
- Adding a check and logging in case `aws` object is not set. 
- 
## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #337, opened 2025-02-21
fix: moving the telemetry logging for initialization options to the l…

suprajaven:
…spRouter

## Problem
Relevant PRs for context : https://github.com/aws/language-server-runtimes/pull/332
https://github.com/aws/language-server-runtimes/pull/310
* Logs/Telemetry is getting invoked for each server, hence it is duplicate and spamming the logs. 

## Solution
* Moving the aws check logic into the router so it does not get invoked each time for each server 
* Removing the logging, can add back if required since it is logging the same thing as the telemetry. 

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/lspRouter.test.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.test.ts
@@ -46,6 +46,7 @@ describe('LspRouter', () => {
         const onInitializeSpy = sandbox.spy(lspConnection, 'onInitialize')
         const onExecuteommandSpy = sandbox.spy(lspConnection, 'onExecuteCommand')
         lspConnection.telemetry.logEvent = sandbox.stub()
+        lspConnection.console.log = sandbox.stub()
 
         lspRouter = new LspRouter(lspConnection, 'AWS LSP Standalone', '1.0.0')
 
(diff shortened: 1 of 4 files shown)


runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -43,6 +43,21 @@ export class LspRouter {
...
+                result: 'Failed',
+                data: {
+                    hasAwsConfig: Boolean(params.initializationOptions?.aws),
+                    logLevel: params.initializationOptions?.logLevel,
+                    initializationOptionsStr: JSON.stringify(params.initializationOptions),
+                },
+                errorData: {
+                    reason: 'aws field is not defined in InitializeResult',
+                },
+            })
```
rahmaniaam, line 58: Missing the logging part?

runtimes/runtimes/lsp/router/lspRouter.test.ts
```diff
@@ -71,6 +72,19 @@ describe('LspRouter', () => {
             assert(lspRouter.clientInitializeParams === initParam)
         })
 
+        it('should log telemetry event when aws config is missing in InitializeParams', async () => {
```
rahmaniaam, line 75: I don't think this covers if there are multiple servers yet, can we add another test for it?

runtimes/runtimes/lsp/router/lspRouter.ts
rahmaniaam: Just read the description, I think it's worth it to keep the logging since it will end up in different places

suprajaven: Ack.

runtimes/runtimes/lsp/router/lspRouter.test.ts
suprajaven: They do not matter here, since we are adding the telemetry before the lspServer initialization - but regardless, I shall add a test with multiple servers present in the params.

runtimes/runtimes/lsp/router/lspRouter.ts
awschristou: It's worth adding the "Unknown initialization error..." log in here. Emitting the telemetry is the most critical thing, but for users who disable telemetry, the log will still provide some sort of clue or signal if users end up sharing their log content when reporting a problem.

rahmaniaam (approved):

PR #338, opened 2025-02-21
feat: add new lsp extension to send dependency information from client to servers

ege0zcan:
## Problem
Servers don't have in depth information regarding dependencies on the client side

## Solution
Define new interface that allows clients to send dependency information to servers 

## License


types/notifyDependencyPaths.ts
```diff
@@ -0,0 +1,10 @@
+export const NOTIFY_DEPENDENCY_PATHS_NOTIFICATION_METHOD = 'aws/notifyDependencyPaths'
+
+export interface NotifyDependencyPathsParams {
+    moduleName: string
+    programmingLangugage: string
```
justinmk3, line 5: typo. Also, "prog lang" may be misleading. This is more like a "mode" or "kind" which triggers heuristics that look at the workspace. In other places (Lambda, SAM CLI, Toolkits, even [CodeWhisperer](https://github.com/aws/aws-toolkit-vscode/blob/cc0455f112c3be8db3232acfb777341f3225c557/packages/core/src/codewhisperer/util/runtimeLanguageContext.ts#L16)), it's called a "runtime".

For example, "node" is not a programming language, but a node project involves js + ts. Which "programming language" would we specify there?

```suggestion
    runtime: string
```

types/notifyDependencyPaths.ts
```diff
@@ -0,0 +1,10 @@
...
+    files: string[]
+    dirs: string[]
```
justinmk3, line 7:
```suggestion
    /** Absolute paths to directories to watch(?) */
    dirs: string[]
```

types/notifyDependencyPaths.ts
```diff
@@ -0,0 +1,10 @@
+export const NOTIFY_DEPENDENCY_PATHS_NOTIFICATION_METHOD = 'aws/notifyDependencyPaths'
```
justinmk3, line 1: (off topic: uppercase is not gaining anything here except a longer, more awkward name.)

types/notifyDependencyPaths.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export interface NotifyDependencyPathsParams {
```
volodkevych, line 3: It's difficult to understand what this protocol is needed for and how to use it. Proper documentation may help.

ege0zcan: +1
I expect some discussions on this PR so I didn't add any documentation just yet since things are fluid at the moment

types/notifyDependencyPaths.ts
ege0zcan: I tried following the existing convention to keep a unified approach but indeed an unwanted side effect is the lengthy constant name. Though there is a clear benefit of having this string in a constant since it will be used by runtimes as well as by clients

types/notifyDependencyPaths.ts
ege0zcan: That's a good point, what about `runtimeLanguage` I still think we need the language information and not the runtime

For a node project with both js and ts files, the client can send one notification for each language, imo this is not redundant but rather being explicit in the communication. I am open to discuss

types/notifyDependencyPaths.ts
ege0zcan: Good point, at least right now I would rather keep that detail out of scope and let the servers decide on depth. It would be easy to add an extra field in the future

types/notifyDependencyPaths.ts
kmile: Also, `notify dependency paths` about what? Is this a change notification that the dependency paths have been updated? If so we can mimic https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWorkspaceFolders and call it

`didChangeDependencyPaths` instead?

kmile:
Thanks for the updates, looks good to me!

kmile (approved):

types/didChangeDependencyPaths.ts
```diff
@@ -0,0 +1,25 @@
+export const DID_CHANGE_DEPENDENCY_PATHS_NOTIFICATION_METHOD = 'aws/didChangeDependencyPaths'
+
+/**
+ * Parameters for notifying AWS language server about dependency paths
+ */
+export interface DidChangeDependencyPathsParams {
+    /** Name of the module being processed */
+    moduleName: string
+    /** Programming language runtime (e.g., 'javascript', 'python', 'java') */
+    runtimeLanguage: string
```
justinmk3, line 10: I'm still not sure that "language" should be specified here. When should the client send "javascript" instead of "typescript"? In that case, the runtime (nodejs) is more relevant, because that also gives a hint about the package manager.

justinmk3: > For a node project with both js and ts files, the client can send one notification for each language, imo this is not redundant but rather being explicit in the communication. I am open to discuss

I thought we discussed that the state for dependency tracking was defined exactly once for a workspace? Thus multiple requests overwrite the previous request.

If that's not the case, then the docstring here should mention "the client must send one notification for each language".

PR #340, opened 2025-02-21
fix: export operational telemetry schemas

natdurlik:
## Problem
We need to share operational telemetry schema definitions between packages.

## Solution
- Dereferenced telemetry schemas into `bundled-telemetry-schema.json`.
- Added bundled schema and generated types to `outDir`.
- Relocated type generation script to runtimes directory for independent compilation.

## License


ege0zcan (approved):

package.json
```diff
@@ -26,7 +26,7 @@
     "scripts": {
         "clean": "ts-node ./script/clean.ts",
         "commitlint": "commitlint --edit",
-        "precompile": "npm run generate-types",
+        "precompile": "npm run generate-types --workspaces --if-present",
```
ege0zcan, line 29: nit: I think `npm --prefix` would be more explicit and potentially slightly more efficient here
 `npm --prefix ./runtimes generate-types`

natdurlik: I can modify the command to `npm run --prefix ./runtimes generate-types`, but this seems just a style change. The `precompile` script in `runtimes/package.json` does not become redundant - it allows us to compile the runtimes package independently without building the entire monorepo.

I could modify the root-level `precompile` to use `npm run precompile --workspaces --if-present`, but this doesn't address the core issue: the generate-types script is still being executed twice unnecessarily.

ege0zcan (approved):

package.json
natdurlik: Yes, npm automatically executes `pre` and `post` hooks for any script, not just built-in commands: https://docs.npmjs.com/cli/v8/using-npm/scripts.

The root level `precompile` is still necessary, as the compile script first runs `tsc --build`, which causes "Cannot find module" errors. Maybe the right solution would be to delete this part from root-level project.json and leave only `npm run compile --workspaces --if-present`, but I didn't research that. I guess we also run `tsc --build` twice because of this.

The root-level `precompile` is currently necessary because `compile` script starts with `tsc --build`, which causes "Cannot find module" errors. We could probably fix this by removing `tsc --build` from root package.json and keeping only `npm run compile --workspaces --if-present`, but I didn't check that. Looks like we also run `tsc --build` twice for each module.

ege0zcan: Thanks for the explanation, I forgot that the generated schemas are used in other parts of the monorepo. Exploring the removal of `tsc --build` command looks like a path worth investigating when we have time, could save few seconds every run

And I still believe changing the root level `precompile` script to `npm run --prefix ./runtimes generate-types` is a small but worthy change since it makes it much simpler to understand in a single glance what the precompile script does compared to going to every subpackage and checking if there is a `precompile` script

natdurlik: Just to clarify, we don't use generated schemas in other parts of the monorepo, failures happen during the `tsc --build` step in the root-level `compile`. As part of this step we build the runtimes project.

Regarding the precompile, I agree, I'll change it to `npm run --prefix ./runtimes generate-types` in a next PR.

PR #351, opened 2025-03-05
fix(runtimes): catch unhandled encryption validation and process exceptions

viktorsaws:
## Problem
On exception, Standalone runtime exits abruptly and NodeJS may print whole application source code in stack trace.

## Solution
Handle auth encryption token validation and process exceptions in Standalone runtime and print specific error message.

## License

### DIFF
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -72,6 +72,11 @@ if (checkAWSConfigFile()) {
     process.env.AWS_SDK_LOAD_CONFIG = '1'
 }
 
+process.on('uncaughtException', err => {
+    console.error('Uncaught Exception:', err.message)
+    process.exit(1)
+})
+
 /**
  * The runtime for standalone LSP-based servers.
  *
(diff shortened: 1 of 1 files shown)


rahmaniaam (approved):

runtimes/runtimes/standalone.ts
```diff
@@ -72,6 +72,11 @@ if (checkAWSConfigFile()) {
     process.env.AWS_SDK_LOAD_CONFIG = '1'
 }
 
+process.on('uncaughtException', err => {
+    console.error('Uncaught Exception:', err.message)
+    process.exit(1)
```
justinmk3, line 77: This nodejs feature is quasi-deprecated, and newer node versions have CLI flags that control how to handle uncaught exceptions/rejections.

- `--unhandled-rejections=mode` https://nodejs.org/api/cli.html#--unhandled-rejectionsmode
- `--abort-on-uncaught-exception` https://nodejs.org/api/cli.html#--abort-on-uncaught-exception

Also note that any code loaded in this nodejs process can overwrite this `on('uncaughtException')` handler (accidentally).

PR #353, opened 2025-03-06
feat(types): types for device code support

hayemaxi:
This PR implements the required types to add device code auth support to the Identity Server. Currently, only PKCE auth flow is supported.

- E_CANCELLED error to indicate the auth flow was cancelled
- `GetSsoTokenProgress` and related types, to indicate to the client whether an authentication is in progress
- `AuthorizationFlowKind` and related types, to indicate what auth flow to use when requesting an SSO token

## License

### DIFF
--- a/runtimes/README.md
+++ b/runtimes/README.md
@@ -211,7 +211,7 @@ Complete Chat parameter and result interfaces can be found in [chat.ts](src/prot
 The Identity Management feature is designed to centralize the management of authentication and identity-related functionality. The APIs consist of:
 
 - Listing and managing user profiles and SSO sessions
-- Obtaining valid SSO access tokens, handling the PKCE login flow as needed
+- Obtaining valid SSO access tokens, handling the PKCE or Device Code login flows as needed
 - Controlling the lifetime and notifications for active SSO tokens
 - Invalidating cached SSO tokens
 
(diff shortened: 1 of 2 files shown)


runtimes/README.md
```diff
@@ -228,6 +228,35 @@ The Identity Management feature is designed to centralize the management of auth
 
 Complete Identity Management parameter and result interfaces can be found in [identity-management.ts](src/protocol/identity-management.ts)
 
+#### Client Capability
```
jpinkney-aws, line 231: 🙌 should this documentation live closer to the language server? Either way thanks for adding this! Hopefully the next person who has to do this won't get as confused as us lol

runtimes/protocol/identity-management.ts
```diff
@@ -31,6 +32,7 @@ export const AwsErrorCodes = {
     E_SSO_TOKEN_SOURCE_NOT_SUPPORTED: 'E_SSO_TOKEN_SOURCE_NOT_SUPPORTED',
     E_TIMEOUT: 'E_TIMEOUT',
     E_UNKNOWN: 'E_UNKNOWN',
+    E_CANCELLED: 'E_CANCELLED',
```
floralph, line 35: question:  Will this be used in conjunction with the CancellationToken?

runtimes/protocol/identity-management.ts
```diff
@@ -151,12 +153,43 @@ export interface IamIdentityCenterSsoTokenSource {
...
+}
+
+export const GetSsoTokenProgressType = new ProgressType<GetSsoTokenProgress>()
+
+// Use this to identify the sendProgress/onProgress call from the identity server.
+// It indicates that an SSO login is currently in progress.
+export const GetSsoTokenProgressToken = 'aws/identity/getSsoToken/progressToken'
+
+export type DeviceCodeAuthorizationFlowKind = 'DeviceCode'
+export type pkceAuthorizationFlowKind = 'PKCE'
```
floralph, line 178: nit: Prefer Pascal-case in name and value for consistency and naming convention.

runtimes/protocol/identity-management.ts
```diff
@@ -151,12 +153,43 @@ export interface IamIdentityCenterSsoTokenSource {
...
+export type AuthorizationFlowKind = DeviceCodeAuthorizationFlowKind | pkceAuthorizationFlowKind
+export const AuthorizationFlowKind = {
+    DeviceCode: 'DeviceCode',
+    PKCE: 'PKCE',
```
floralph, line 182: nit: Prefer Pascal-case in name and value as naming convention.

runtimes/protocol/identity-management.ts
```diff
@@ -151,12 +153,43 @@ export interface IamIdentityCenterSsoTokenSource {
...
+export type Complete = 'Complete'
+
+export type GetSsoTokenProgressState = InProgress | Complete
+
+export const GetSsoTokenProgressState = {
+    InProgress: 'InProgress',
+    Complete: 'Complete',
+} as const
+
+export interface GetSsoTokenProgress {
```
floralph, line 166: question:  Is this needed on the PKCE flow too?  Given only the InProgress and Complete states, does the client not already know this by calling GetSsoToken and awaiting a response?

floralph (approved):

runtimes/protocol/identity-management.ts
hayemaxi: Yes

runtimes/protocol/identity-management.ts
hayemaxi: Yes but server-side is more granular, e.g. the `message` field containing information, or only send put a progress bar if waiting for the user externally, e.g. to finish the sign-in process at the provided link.

PKCE could also implement a progress bar for consistency, but I would say it is not as important as device code where we can use this as another way to show the code to the users.

runtimes/README.md
hayemaxi: IMO yes but it seems that the convention is to document it in this repo instead. To avoid fragmented docs we should leave it here for now.

pras0131 (approved):

PR #355, opened 2025-03-07
Update protocol for regions use cases

volodkevych:
## Problem

There is no way to request chat UI to open a new tab from server. This will be needed to restore chat state after profile switch.

Given connection type becomes required now for region detection, we should move it to `aws/credentails/token/update` request - it simplifies the flow and less error-prone.

## Solution

- extended chat protocol with `'aws/chat/openTab'`
- extended `UpdateCredentialsParams` in `'aws/credentials/token/update'` with `startUrl`
- moved the logic to define connection type to runtimes

## License

### DIFF
--- a/runtimes/runtimes/auth/auth.ts
+++ b/runtimes/runtimes/auth/auth.ts
@@ -18,6 +18,8 @@ import {
     CredentialsProvider,
 } from '../../server-interface'
 
+export const BUILDER_ID_START_URL = 'https://view.awsapps.com/start'
+
 export function isIamCredentials(credentials: Credentials): credentials is IamCredentials {
     const iamCredentials = credentials as IamCredentials
     return iamCredentials?.accessKeyId !== undefined && iamCredentials?.secretAccessKey !== undefined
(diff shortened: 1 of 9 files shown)


runtimes/runtimes/auth/auth.ts
```diff
@@ -137,6 +143,18 @@ export class Auth {
         })
     }
 
+    private async handleBearerCredentialsMetadata(metadata?: ConnectionMetadata) {
+        if (metadata) {
+            this.connectionMetadata = metadata
```
viktorsaws, line 148: Guess would be good to do more validation of incoming `metadata` object, since client could send whatever. Also not sure if we should return an error to client here, but at least logging something would be good.

runtimes/runtimes/auth/auth.ts
```diff
@@ -70,6 +72,11 @@ export class Auth {
             getConnectionMetadata: () => {
                 return this.connectionMetadata
             },
+
+            getConnectionType: () => {
+                const startUrl = this.connectionMetadata?.sso?.startUrl
+                return !startUrl ? 'none' : startUrl.includes(BUILDER_ID_START_URL) ? 'builderId' : 'identityCenter'
```
viktorsaws, line 78: nit: missing unit test for when `getConnectionType` returns `none`

viktorsaws (approved):

runtimes/runtimes/auth/auth.ts
volodkevych: All the fields in `metadata` are optional. Given that I'm not sure I should validate any other fields. If client send the metadata, he defined it already, even if it's empty.

runtimes/runtimes/auth/auth.ts
volodkevych: added it

viktorsaws (approved):

PR #361, opened 2025-03-11
chore: adding keepAlive options to AWS SDK client

pras0131:
## Problem
Adding http keepalive agent in AWS SDK Client to help in reducing the latency of generateCompletions API.

## License

### DIFF
--- a/runtimes/runtimes/util/standalone/proxyUtil.ts
+++ b/runtimes/runtimes/util/standalone/proxyUtil.ts
@@ -145,6 +145,9 @@ export class ProxyConfigManager {
         const agentOptions = {
             ca: certs,
             rejectUnauthorized: true, // Always validate certificates
+            keepAlive: true,
+            keepAliveMsecs: 30000, // Keep alive for 30 seconds
+            maxSockets: 10, // Maximum number of sockets to allow per host
         }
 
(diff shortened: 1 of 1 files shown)


ege0zcan (approved):
Do we know the implications of this? 

I understand this helps with latency on cwspr endpoints but does it impact all aws sdk calls? Is there a way we can test this?

pras0131:
> Do we know the implications of this?
> 
> I understand this helps with latency on cwspr endpoints but does it impact all aws sdk calls? Is there a way we can test this?

I did test it when I made a similar change for language-servers. Confirmed that it's fine to do it for aws sdk calls.

justinmk3:
This is a "fix" or "feat", not "chore". People really over-use the "chore" marker, which makes PR titles meaningless and defeats the purpose of conventional commits.

pras0131:
> This is a "fix" or "feat", not "chore". People really over-use the "chore" marker, which makes PR titles meaningless and defeats the purpose of conventional commits.

I don't think we can call it a fix as it just introduces additional config, nor it is a feature. That's why chore is used here.

justinmk3:
Config changes can (and this one does) have effects on the application. "chore" (and "refactor") implies a non-functional change. This is a functional change.

PR #364, opened 2025-03-12
feat(types): add next token to inline completion interface

zixlin7:
## Problem
To support pagination for inline recommendations we need to add nextToken for both request and response types
## Solution
update types for `InlineCompletionWithReferencesParam` and `InlineCompletionListWithReferences`

## License


pras0131 (approved):

runtimes/protocol/inlineCompletionWithReferences.ts
```diff
@@ -8,7 +8,9 @@ import {
     InlineCompletionParams,
 } from './lsp'
 
-export type InlineCompletionWithReferencesParams = InlineCompletionParams & {}
+export type InlineCompletionWithReferencesParams = InlineCompletionParams & {
+    nextToken?: string
```
viktorsaws, line 12: Instead of adding custom `nextToken` parameter to protocol, it is better to use PartialResult params with `partialResultToken`, as it is standard way in LSP to signal to client that more results are available on request 
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#partialResultParams

viktorsaws (changes requested):

viktorsaws (approved):

jpinkney-aws (approved):

PR #366, opened 2025-03-13
feat: improvements for "openTab" request flow

volodkevych:
## Problem

- There is no mechanism to return error from `chat-client` to the calling party

## Solution

Extended the protocol to support errors as part of  response over `postMessage`.

## License

### DIFF
--- a/chat-client-ui-types/src/uiContracts.ts
+++ b/chat-client-ui-types/src/uiContracts.ts
@@ -3,6 +3,8 @@ import {
     ChatOptions,
     CodeSelectionType,
     ReferenceTrackerInformation,
+    OPEN_TAB_REQUEST_METHOD,
+    OpenTabResult,
 } from '@aws/language-server-runtimes-types'
 export { InsertToCursorPositionParams } from '@aws/language-server-runtimes-types'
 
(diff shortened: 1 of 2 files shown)


chat-client-ui-types/src/uiContracts.ts
```diff
@@ -117,3 +117,8 @@ export interface CopyCodeToClipboardMessage {
     command: typeof COPY_TO_CLIPBOARD
     params: CopyCodeToClipboardParams
 }
+
+export interface ErrorResult {
+    message: string
+    type: 'InvalidRequest' | 'InternalError' | 'UnknownError' | string
+}
```
justinmk3, line 124: Looks interesting, is this intended for general use, or only opentab?

Would it make sense for this be modeled as `Result` (and include 'Success')), instead of `ErrorResult`?

types/chat.ts
```diff
@@ -195,7 +195,12 @@ export interface FollowUpClickParams {
...
 /*
     Defines parameters for opening a tab.
-    Opens existing tab if `tabId` is provided, otherwise creates a new tab and opens it.
+    Opens existing tab if `tabId` is provided, otherwise creates a new tab
+    with options provided in `options` parameter and opens it.
 */
-export interface OpenTabParams extends Partial<TabEventParams> {}
+export interface OpenTabParams extends Partial<TabEventParams> {
+    options?: {
+        needWelcomeMessages?: boolean
```
dogusata, line 203: I don't oppose to the idea `needWelcomeMessages` but don't like the idea having these messages hardcoded. And there is a case, when user selects some text and trigger's Explain (or similar) from Q right click context menu. It should open a new tab with the selected code block as a user message. Shouldn't this options have `messages[]` also? Then `needWelcomeMessages` might be not necessary anymore.

volodkevych: I thought these welcome messages were hardcoded on mynah-ui, but I see now that they are hardcoded in `chat-client`. Then, it makes sense to me we should not use them as universal welcome massages for any tab. I'll rollback this.

chat-client-ui-types/src/uiContracts.ts
volodkevych: It can be for general use, but handling responses through `postMessage` would require improvement in general.

ege0zcan (approved):
LGTM!

PR #376, opened 2025-03-17
feat(runtimes): extend awsClientCapabilities generically to signal server-specific capabilities

patrickvbeurden:
## Problem

For language server `aws-lsp-codewhisperer` in particular, we need a way for clients to be able to signal that they support Q developer profiles at initialization.

## Solution

Extend the `awsClientCapabilities` entry of `AWSInitializationOptions` with `[key: string]: any`. This enables clients to signal support for server-specific capabilities in a generic way.

Testing: build and installed the change into a local copy of the language servers repository to verify that the server can now parse for any client capabilities.

## License

### DIFF
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -107,6 +107,7 @@ export interface AWSInitializationOptions {
      * The client signals AWS capabilities it supports.
      */
     awsClientCapabilities?: {
+        [key: string]: any
         window?: {
             notifications?: boolean
         }


runtimes/protocol/lsp.ts
```diff
@@ -107,6 +107,7 @@ export interface AWSInitializationOptions {
      * The client signals AWS capabilities it supports.
      */
     awsClientCapabilities?: {
+        [key: string]: any
```
volodkevych, line 110: How will clients know what is supported?

viktorsaws: this will need to be documented by each individual Server implementation. I don't think we need to define then in Runtimes protocol, since it is generic. In this sense, it is not different from LSP Configuration or Commands supported by individual Servers.

viktorsaws (approved):

runtimes/protocol/lsp.ts
volodkevych: It will be used by all Q-related servers, not some specific one. I think we need to have this more explicit.

viktorsaws: But Q-related servers are quite specific servers. We don't encode `aws.q` configuration sections in protocol, I don't think we should start encoding options, as they are used same way.

patrickvbeurden: Me and Viktor agreed to merge this pr to unblock ourselves, we can later come back and follow up to make it more specific if we want.

PR #378, opened 2025-03-18
feat(runtimes): cache and expose initializeParams as a feature

patrickvbeurden:
## Problem

It is becoming more frequent that the `initializeParams` sent by the client have to be referenced in different parts of a server _after_ initialization. We need a more scalable way to retrieve these parameters to avoid duplication or having to couple components with the initialization code.

## Solution

The `LspRouter` already caches the client `initializeParams`, so to make these available we can extend the `lsp` feature with a method/property `getClientInitializeParams`, which retrieves the cached client params in the `LspRouter`. Example usage:

```ts
// somewhere in language-servers/servers/aws-lsp-*
const params: InitializeParams = lsp.getClientInitializeParams()
```

Testing: Created a reusable handler for this purpose with unit test. Also packed and installed in a local copy of `language-servers` to verify this feature is now available.

## License

### DIFF
--- a/runtimes/runtimes/base-runtime.ts
+++ b/runtimes/runtimes/base-runtime.ts
@@ -68,6 +68,7 @@ import { WebBase64Encoding } from './encoding'
 import { LoggingServer } from './lsp/router/loggingServer'
 import { Service } from 'aws-sdk'
 import { ServiceConfigurationOptions } from 'aws-sdk/lib/service'
+import { getClientInitializeParamsHandlerFactory } from './util/lspCacheUtil'
 
 declare const self: WindowOrWorkerGlobalScope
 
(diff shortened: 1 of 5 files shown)


viktorsaws (approved):

runtimes/runtimes/base-runtime.ts
```diff
@@ -216,6 +217,9 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
                     lspConnection.onNotification(didChangeDependencyPathsNotificationType, handler)
                 },
             },
+            cache: {
```
viktorsaws, line 220: nit: I'd just move `getClientInitializeParams` method to `lsp` object itself and drop nested `cache` object.

viktorsaws (approved):

PR #379, opened 2025-03-18
chore: notify servers of bearer token signout so that servers can update/reset local state

pras0131:
## Problem
We need to notify servers in case of bearer token sign out so that they can update or reset the local state, which means resetting the developer profile. It's needed when profile has changed, but server has not received updated profile yet and calls are made to the wrong region (race condition)

## Solution
Introduced a change so that Auth utility has access to the router class in order to be able to send messages/notifications to the servers. Without doing so, we only have reference to the lspConnection only, but not the servers.

## License

### DIFF
--- a/runtimes/runtimes/auth/auth.ts
+++ b/runtimes/runtimes/auth/auth.ts
@@ -1,5 +1,5 @@
 import { jwtDecrypt } from 'jose'
-import { Connection } from 'vscode-languageserver'
+import { CancellationTokenSource, Connection, ExecuteCommandParams } from 'vscode-languageserver'
 import { CredentialsEncoding } from './standalone/encryption'
 import {
     UpdateCredentialsParams,
@@ -17,6 +17,7 @@ import {
     CredentialsType,
     CredentialsProvider,
 } from '../../server-interface'
+import { LspRouter } from '../lsp/router/lspRouter'
 
 export const BUILDER_ID_START_URL = 'https://view.awsapps.com/start'
 
@@ -37,6 +38,7 @@ export class Auth {
 
     private key: Buffer | undefined
     private credentialsEncoding: CredentialsEncoding | undefined
+    private lspRouter!: LspRouter
 
     constructor(
         private readonly connection: Connection,
@@ -80,6 +82,10 @@ export class Auth {
         }
 
         this.registerLspCredentialsUpdateHandlers()
+    }
+
+    public setLspRouter(lspRouter: LspRouter) {
+        this.lspRouter = lspRouter
     }
 
     public getCredentialsProvider(): CredentialsProvider {
@@ -139,6 +145,18 @@ export class Auth {
         this.connection.onNotification(bearerCredentialsDeleteNotificationType, () => {
             this.bearerCredentials = undefined
             this.connectionMetadata = undefined
+            let params: ExecuteCommandParams = {
+                command: 'bearerCredentialsDeleteCommand',
+            }
+            const tokenSource = new CancellationTokenSource()
+            this.lspRouter
+                ?.executeCommand(params, tokenSource.token)
+                .then(result => {
+                    this.connection.console.info(`Server notified of bearer token deletion`)
+                })
+                .catch(error => {
+                    this.connection.console.error(`Error while notifying server of bearer token update, ${error}`)
+                })
             this.connection.console.info('Runtime: Deleted bearer credentials')
         })
     }
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -422,6 +422,8 @@ export const standalone = (props: RuntimeProps) => {
             })
         })
 
+        auth.setLspRouter(lspRouter)
+
         // Free up any resources or threads used by Servers
         lspConnection.onExit(() => {
             disposables.forEach(d => d())


runtimes/runtimes/auth/auth.ts
```diff
@@ -139,6 +145,18 @@ export class Auth {
         this.connection.onNotification(bearerCredentialsDeleteNotificationType, () => {
             this.bearerCredentials = undefined
             this.connectionMetadata = undefined
+            let params: ExecuteCommandParams = {
+                command: 'bearerCredentialsDeleteCommand',
```
volodkevych, line 149: We have this command already: `aws/credentials/token/remove` method. Why do we need new one?

viktorsaws: I guess this is needed to notify all registered Servers, that `bearerCredentialsDeleteNotification` was sent, because currently they can not subscribe to it. But I don't think that we should be using LSP command for this. Can we expose some event `credentialsProvider` on object, that is injected to every Server already?

We could extend `credentialsProvider` with a callback handler, similar like we do with `lsp` object in runtimes https://github.com/aws/language-server-runtimes/blob/e3d4e1772d29955ae71a6761f1be3251c443215e/runtimes/runtimes/standalone.ts#L305-L306

volodkevych: Cannot we route `aws/credentials/token/remove` to runtime and servers?

runtimes/server-interface/auth.ts
```diff
@@ -12,4 +12,5 @@ export interface CredentialsProvider {
     getCredentials: (type: CredentialsType) => Credentials | undefined
     getConnectionMetadata: () => ConnectionMetadata | undefined
     getConnectionType: () => SsoConnectionType
+    onBearerCredentialsDelete: (handler: () => void) => void
```
volodkevych, line 15: Can we generalize this to `onCredentialsDeleted`?

runtimes/runtimes/auth/auth.ts
```diff
@@ -72,16 +75,22 @@ export class Auth {
...
             },
+            onBearerCredentialsDelete: (handler: () => void) => {
+                this.bearerCredentialsDeleteHandler = handler
+            },
         }
 
         this.registerLspCredentialsUpdateHandlers()
     }
 
+    public setLspRouter(lspRouter: LspRouter) {
```
volodkevych, line 90: We should pass `lspRouter` in ctor. Setting it like this is more fragile.

runtimes/runtimes/auth/auth.ts
```diff
@@ -139,6 +148,7 @@ export class Auth {
         this.connection.onNotification(bearerCredentialsDeleteNotificationType, () => {
             this.bearerCredentials = undefined
             this.connectionMetadata = undefined
+            this.lspRouter.onBearerTokenDeletion()
```
volodkevych, line 151: We introduce a new pattern of using router here. Now, not only events from LSP client are routed, but also other server-interface events. Which can be ok actually, if it simplifies the implementation.

runtimes/server-interface/auth.ts
pras0131: we can do that but here we were doing all these things for bearer token, so I though naming it explicitly would be better.

runtimes/runtimes/auth/auth.ts
pras0131: We init [auth in the beginning](https://github.com/aws/language-server-runtimes/blob/e3d4e1772d29955ae71a6761f1be3251c443215e/runtimes/runtimes/standalone.ts#L136) and lspRouter later. did not want to change the order,so I created a different setter rather than adding to constructor

runtimes/server-interface/auth.ts
pras0131: but I agree that making it generic would be better. I can make this change.

runtimes/runtimes/standalone.ts
```diff
@@ -408,6 +407,10 @@ export const standalone = (props: RuntimeProps) => {
                 }
             )
 
+            if (credentialsProvider) {
```
viktorsaws, line 410: credentialsProvider should never be undefined, do we need this condition?

runtimes/runtimes/standalone.ts
```diff
@@ -408,6 +407,10 @@ export const standalone = (props: RuntimeProps) => {
...
+                credentialsProvider.onCredentialsDeleted = lspServer.setCredentialsDeleteHandler
```
viktorsaws, line 411: Looks like same wiring in `base-runtime` is missing, we don't attach `onCredentialsDeleted` handler there

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -162,6 +162,10 @@ ${JSON.stringify({ ...result.capabilities, ...result.awsServerCapabilities })}`
         )
     }
 
+    onCredentialsDeletion = (type: CredentialsType): void => {
+        this.servers.forEach(server => server.notifyCredentialsDeletion(type))
```
viktorsaws, line 166: nit: could use `routeNotificationToAllServers` method below for this

runtimes/runtimes/auth/auth.ts
```diff
@@ -139,6 +145,7 @@ export class Auth {
         this.connection.onNotification(bearerCredentialsDeleteNotificationType, () => {
             this.bearerCredentials = undefined
             this.connectionMetadata = undefined
+            this.lspRouter.onCredentialsDeletion('bearer')
```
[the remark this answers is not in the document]
pras0131: > Should we call onCredentialsDeletion in iamCredentialsDeleteNotificationType handler as well?
Right now as we only made changes to the service manager for bearer token, so intentionally missed it.

Can add unit test & above if required as a follow up PR.

viktorsaws (approved):

PR #368, opened 2025-03-14
feat: extended protocol for conversation-based agents experience support

volodkevych:
## Problem

We need to extend chat protocol to support agent-specific flow.

## Solution

Protocol and server interface were extended to support:
- `'aws/chat/sendChatUpdate'` - for pushing async updates from server to client
- `'aws/chat/fileClick'` - for opening file or executing related file action, if provided
- `'aws/selectWorkspaceItem'` - to give possibility to prompt customer select another workspace folder in IDE if current is too big
- `'aws/openFileDiff'` - to trigger open file differences in editor

`ChatOptions` `QuickActionCommand` were extended:
- `disabled` was removed - disabling quick actions within quick action tabs needs to be re-modelled. Current approach will not work as chat options from different servers are merged in Flare runtimes. No real client is using this field now.

`OpenTabParams` params were extended to give possibility to configure contents and state of the tab if a new tab is requested.

Other minor changes:
`fileList` and `prompts` supported in chat message responses.
`placeholderText` can be set.

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #387, opened 2025-03-20
docs: conversation-based agents

justinmk3:
## Problem

conversational chat protocol added in https://github.com/aws/language-server-runtimes/pull/368 had some comments that give practical info about the protocol.

## Solution

add the info to docstrings.

## License

### DIFF
--- a/types/workspace.ts
+++ b/types/workspace.ts
@@ -1,5 +1,6 @@
 import { URI, TextDocument } from './lsp'
 
+/** Server to client. Used to ask the user to select a folder, if the workspace folder is too big */
 export const SELECT_WORKSPACE_ITEM_REQUEST_METHOD = 'aws/selectWorkspaceItem'
 export const OPEN_FILE_DIFF_NOTIFICATION_METHOD = 'aws/openFileDiff'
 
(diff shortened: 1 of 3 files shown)


types/workspace.ts
```diff
@@ -1,5 +1,6 @@
 import { URI, TextDocument } from './lsp'
 
+/** Server to client. Used to ask the user to select a folder, if the workspace folder is too big */
```
volodkevych, line 3: "Server to client" is specified in README - I would prefer to keep it there, specifying the direction of request in one place only looks inconsistent to me.
"if the workspace folder is too big" - this is just one of examples when it will be used, I do not think we need it in comment like this.
"select a folder" -> "select files(s) or folder(s)"

types/workspace.ts
```diff
@@ -18,7 +19,9 @@ export interface SelectWorkspaceItemResult {
 }
 
 export interface OpenFileDiffParams {
+    /** The initial version of the file. */
     originalFileUri: URI
+    /** Indicates if the original file was deleted and fileContent contains the new/updated file contents. */
```
volodkevych, line 24: can we please spilt it into 2 comments for 2 fields?

PR #390, opened 2025-03-27
feat: add license check workflow

suprajaven:
## Problem
* Adding a license check github workflow to run on every PR. 

## Solution

* Introducing a github workflow that will check if packages belong to a list of pre approved licenses. 

## License

### DIFF
--- a/.github/workflows/license-check.yaml
+++ b/.github/workflows/license-check.yaml
@@ -3,7 +3,6 @@ name: License Check
 on:
   pull_request:
     branches: [main]
-    types: [opened, reopened, ready_for_review, synchronize]
 
 jobs:
   license-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/license-check.yaml
```diff
@@ -0,0 +1,52 @@
+name: License Check
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, reopened, ready_for_review, synchronize]
```
natdurlik, line 6: nit: We can have this check on all types of `pull_request`, so we don't need to specify `types` field at all. What do you think?

.github/workflows/license-check.yaml
```diff
@@ -0,0 +1,52 @@
...
+                echo "License for package '$package_name' is either not pre-approved or package is not in the excluded package list."
+                exit 1
+              fi
+            fi
+          done
+
+          IFS=$' \t\n'
+        }
+
+        LICENSE_CHECK_RESULT=$(license-checker --production --exclude $EXCLUDED_LICENSES)
```
natdurlik, line 51: nit: I think we could also use `--excludePackages` flag instead of for loop?

.github/workflows/license-check.yaml
```diff
@@ -0,0 +1,52 @@
...
+      run: |
+        npm install -g license-checker
+
+    - name: Run license check
+      run: |
+        #!/bin/bash
+
+        set -euo pipefail
+
+        EXCLUDED_PACKAGES=("@amzn/eevee-service" "@aws/lsp-codewhisperer" "@aws/lsp-core" "@c9/eevee-scanner" "@amzn/dexp-runtime-server-build-configuration")
```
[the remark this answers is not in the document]
suprajaven: ack

.github/workflows/license-check.yaml
suprajaven: Agreed

.github/workflows/license-check.yaml
```diff
@@ -0,0 +1,30 @@
+name: License Check
```
natdurlik, line 1: Will we add this workflow to the language-servers repo too?

natdurlik (approved):

PR #392, opened 2025-03-28
feat: collect telemetry data from standalone runtime

natdurlik:
## Problem
We want to be able to collect telemetry data about server crashes, errors, resource usages.

## Solution
* Implemented telemetry data collection: https://github.com/aws/language-server-runtimes/pull/392/commits/f533827d4aedf6a4c884f5ea91db2149df57e43c
* Enforced compile-time type checking from generated schemas https://github.com/aws/language-server-runtimes/pull/392/commits/eb5664750282cc3afac5ea96576dfbf27370bda9 and https://github.com/aws/language-server-runtimes/pull/392/commits/86b3a7fd896120dbf92c32bbca94100db57921a4

## License


runtimes/runtimes/util/telemetryLspServer.ts
```diff
@@ -12,6 +12,32 @@ const DEFAULT_TELEMETRY_GATEWAY_ENDPOINT = ''
...
+    optel.registerGaugeProvider('ResourceUsageMetric', () => process.memoryUsage().heapUsed, { type: 'heapUsed' })
+    optel.registerGaugeProvider('ResourceUsageMetric', () => process.memoryUsage().heapTotal, { type: 'heapTotal' })
+    optel.registerGaugeProvider('ResourceUsageMetric', () => process.memoryUsage().rss, { type: 'rss' })
+}
+
+function setServerCrashTelemetryListeners() {
+    const optel = OperationalTelemetryProvider.getTelemetryForScope(getRuntimeScopeName())
+
+    // Handles both 'uncaughtException' and 'unhandledRejection'
+    process.on('uncaughtExceptionMonitor', async (err, origin) => {
```
viktorsaws, line 30: We need to record `unhandledRejection` for not caught Promise rejections

natdurlik: Was this not working properly before? The nodejs docs https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor say the `origin` field should tell us if it came from an unhandled rejection or from synchronous errors.

runtimes/runtimes/operational-telemetry/aws-metrics-exporter.ts
```diff
@@ -105,16 +105,20 @@ export class AwsMetricExporter implements PushMetricExporter {
...
+            },
         }
 
         const isValid = this.eventValidator.validateEvent(result)
         if (!isValid) {
             throw Error(`Invalid operational event: ${result}`)
         }
 
-        return result as OperationalEvent
+        return result as any
```
volodkevych, line 122: Why was this change needed?

runtimes/runtimes/auth/auth.ts
```diff
@@ -207,6 +213,12 @@ export class Auth {
             this.connection.console.info(
                 `Runtime: Failed to update Connection metadata with error: ${error?.message || 'unknown'}`
             )
+            OperationalTelemetryProvider.getTelemetryForScope(TELEMETRY_SCOPES.RUNTIMES).recordEvent('ErrorEvent', {
+                errorType: 'caughtError',
+                errorName: error?.name ?? 'unknown',
+                errorCode: error?.code ?? '',
+                message: 'Failed to update Connection metadata',
```
volodkevych, line 220: Will we be filtering specific errors by `message` if we need alarm specific to an error type? Why don't we use `errorType` for that?

runtimes/runtimes/util/telemetryLspServer.ts
```diff
@@ -53,6 +79,7 @@ export function getTelemetryLspServer(
 
         if (typeof optOut === 'boolean') {
             OperationalTelemetryProvider.getTelemetryForScope('').toggleOptOut(optOut)
+            setMemoryUsageTelemetry()
```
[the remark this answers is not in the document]
natdurlik: When telemetry is re-enabled after being disabled, memory usage collection needs to be registered again.

runtimes/runtimes/operational-telemetry/aws-metrics-exporter.ts
natdurlik: We no longer accept `[k: string]: unknown;` parameters in the generated type, as a result of adding `"additionalProperties": false` to JSON schemas. When we fill the `result` event from a `Record<string, number> `object, TypeScript cannot verify its correctness at compile time.

runtimes/runtimes/auth/auth.ts
natdurlik: Could you clarify what you mean by error type? We can use errorName to differentiate between different errors classes in Nodejs, such as EvalError, RangeError, SyntaxError. Are you suggesting we create our own custom error types rather than just custom error messages?

volodkevych: Yes, I think we will need more concrete error types rather than 'caughtError' or 'uncaughtError'. Or another field that would give possibility to provide error reason. Otherwise it can be difficult to identify that some specific error started happening more often. For example, jump in errors thrown on init flow can be unnoticed among errors thrown on inline completion request, because rate of those errors will be completely different.

runtimes/runtimes/chat/encryptedChat.ts
```diff
@@ -67,8 +73,17 @@ export class EncryptedChat extends BaseChat {
                     let decryptedRequest
                     try {
                         decryptedRequest = await this.decodeRequest<DecryptedRequestType>(request)
-                    } catch (err: unknown) {
+                    } catch (err: any) {
                         let errorMessage = 'Request could not be decrypted'
+                        OperationalTelemetryProvider.getTelemetryForScope(TELEMETRY_SCOPES.RUNTIMES).recordEvent(
```
ege0zcan, line 78: I think we should move line 87 above the getTelemetryForScope call, so it should be:

```
let errorMessage = 'Request could not be decrypted'
if (err instanceof Error) errorMessage = err.message
OperationalTelemetryProvider.getTelemetryForScope(TELEMETRY_SCOPES.RUNTIMES).recordEvent(...
```

runtimes/runtimes/util/standalone/certificatesReaders.ts
```diff
@@ -57,7 +58,14 @@ export function readLinuxCertificates(): string[] {
 
     // Step 3: Handle errors and return results
     if (!hasSeenCertificate && firstError) {
-        console.log('Error when reading Linux certificates')
+        const errorMessage = 'Error when reading Linux certificates'
+        console.log(errorMessage)
+        OperationalTelemetryProvider.getTelemetryForScope(TELEMETRY_SCOPES.RUNTIMES).recordEvent('ErrorEvent', {
+            errorType: 'caughtError',
+            errorName: firstError?.name ?? 'unknown',
+            errorCode: '',
```
ege0zcan, line 66: Is it possible `firstError` object has a code or a message? If so would we want to use those fields?

runtimes/runtimes/operational-telemetry/operational-event-validator.test.ts
```diff
@@ -48,8 +46,14 @@ describe('OperationalEventValidator', () => {
...
             const invalidEvent = {
-                name: 'ServerCrashEvent',
-                crashType: 'Error',
+                baseInfo: {
+                    name: 'ErrorEvent',
+                    timestamp: 100000,
+                },
+                errorAttr: {
+                    errorName: 'Error',
+                    errorType: 'uncoughtException',
```
ege0zcan, line 55: nit: typo

runtimes/runtimes/auth/auth.ts
ege0zcan: I would also like to understand what we expect from the `errorType` field. I can see the value in grouping all the errors that we were able to catch and process, but on the other side it looks like most of the error events will go under this category which might make the errorType field not very useful. 

As an alternative, we can consider using the errorType field to create our own error types to make filtering simpler. We could have `connectionError`, `encryptionError`, `proxyError` etc. 


Also nit: the first letter in `Connection` shouldn't be capitalized

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
```diff
@@ -99,16 +99,20 @@ export class AwsSpanExporter implements SpanExporter {
 
     private spanToOperationalEvent(span: ReadableSpan): OperationalEvent {
         const unixEpochSeconds = span.endTime[0]
-        const result: Record<string, any> = {}
-
-        result['name'] = span.name
-        result['timestamp'] = unixEpochSeconds
-        for (const key of Object.keys(span.attributes)) {
-            result[key] = String(span.attributes[key])
+        const event = JSON.parse(span.attributes['event.attributes'] as string)
```
ege0zcan, line 102: Just want to double check that here it is guaranteed that the span object will be a proper and valid JSON right? 

JSON.parse can fail if the provided string is invalid json and I want to make sure we don't have a case where our telemetry collection throws an error

runtimes/runtimes/operational-telemetry/operational-telemetry.ts
```diff
@@ -65,3 +73,7 @@ export class OperationalTelemetryProvider {
         return new ScopedTelemetryService(scopeName, OperationalTelemetryProvider.telemetryInstance)
     }
 }
+
+export enum TELEMETRY_SCOPES {
```
ege0zcan, line 77: Enums are a bit weird, potentially we could do:

```
export const TELEMETRY_SCOPES = {
    RUNTIMES: 'language-server-runtimes',
} as const;
```

or 
```
export type TELEMETRY_SCOPE = 'language-server-runtimes';
```

runtimes/runtimes/operational-telemetry/telemetry-schemas/resource-usage-metric-schema.json
```diff
@@ -4,29 +4,41 @@
...
+                    "$ref": "./base-event-schema.json"
+                },
+                "resourceValues": {
+                    "type": "object",
+                    "$ref": "#/definitions/ResourceUsageAttr"
                 }
-            ]
+            }
+        },
+        "ResourceUsageAttr": {
```
ege0zcan, line 20: nit: why are we shortening the names here, I personally would prefer if Attr was more explicit, does it stand for Attributes?

runtimes/runtimes/util/telemetryLspServer.ts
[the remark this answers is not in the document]
ege0zcan: In that case should we also call `setServerCrashTelemetryListeners` again or is registering once enough?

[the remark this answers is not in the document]
natdurlik: `setMemoryUsageTelemetry` registers handlers in the opentelemetry registry, requiring re-registration after opentelemetry is turned off. In contrast, `setServerCrashTelemetryListeners` registers handlers in the Node.js process, so it only needs to be done once.

runtimes/runtimes/chat/encryptedChat.ts
natdurlik: I chose not to use err.message because we don't have control over its content. It potentially could contain PII.

runtimes/runtimes/operational-telemetry/operational-event-validator.test.ts
natdurlik: The test is correctly catching this mistake 😄

runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts
natdurlik: I agree this is important. We have try-catch block for all errors that could occur during extraction and sending in the `export` function: https://github.com/aws/language-server-runtimes/blob/collect-telemetry-from-standalone/runtimes/runtimes/operational-telemetry/aws-spans-exporter.ts#L31. 

JSON.parse should work here since we use the corresponding JSON.stringify(eventAttr) to set this attribute: https://github.com/aws/language-server-runtimes/blob/collect-telemetry-from-standalone/runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts#L164.

runtimes/runtimes/standalone.ts
```diff
@@ -81,9 +81,8 @@ if (checkAWSConfigFile()) {
     process.env.AWS_SDK_LOAD_CONFIG = '1'
 }
 
-process.on('uncaughtException', err => {
```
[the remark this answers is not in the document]
natdurlik: I left the `console.error` message in place because I thought we want to print it as well. As far as I understand we can register multiple 'uncaughtExceptionMonitor' handlers.

runtimes/runtimes/operational-telemetry/telemetry-schemas/error-event-schema.json
```diff
@@ -0,0 +1,45 @@
...
+                    "type": "string"
+                },
+                "errorType": {
+                    "type": "string",
+                    "enum": ["caughtError", "uncaughtException", "unhandledRejection", "other"]
+                },
+                "errorCode": {
+                    "type": "string"
+                },
+                "errorMessage": {
```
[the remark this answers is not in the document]
natdurlik: My initial idea was to distinguish between two messages: `errorMessage` for `error.message` and `message` for our own custom one. I will delete it for now.

runtimes/runtimes/operational-telemetry/telemetry-schemas/resource-usage-metric-schema.json
natdurlik: agree, renamed to Attributes.

runtimes/runtimes/chat/encryptedChat.ts
ege0zcan: The error comes from `decodeRequest` and I see that we have different error messages specified for different types of failures during the decode. If we use `Request could not be decrypted` for every error, we wouldn't have the granularity. And even if the underlying `jwtDecrypt` event throws an error it looks like the error messages are usually straightforward strings that wouldn't contain PII: https://github.com/panva/jose/blob/fa04d88f3532a752f5a1b8a71e628494d13bb07e/src/jwt/decrypt.ts#L80


```
    private async decodeRequest<T>(request: EncryptedRequestParams): Promise<T> {
        if (!this.key) {
            throw new Error('No encryption key')
        }

        if (this.encoding === 'JWT') {
            const result = await jwtDecrypt(request.message, this.keyBuffer, {
                clockTolerance: 60,
                contentEncryptionAlgorithms: [CONTENT_ENCRYPTION_ALGORITHM],
                keyManagementAlgorithms: [KEY_MANAGEMENT_ALGORITHM],
            })

            if (!result.payload) {
                throw new Error('JWT payload not found')
            }
            return result.payload as T
        }
        throw new Error('Encoding mode not implemented')
    }
   ```

File: runtimes/runtimes/standalone.ts (hunk 11 above)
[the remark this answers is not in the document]
Comment (ege0zcan) [reply runtimes/runtimes/standalone.ts:85]: I see, that makes sense, thanks for the clarification

File: runtimes/runtimes/operational-telemetry/telemetry-schemas/error-event-schema.json [hunk 13]
```diff
@@ -0,0 +1,45 @@
...
+            }
+        },
+        "ErrorEventAttributes": {
+            "title": "ErrorEventAttributes",
+            "type": "object",
+            "additionalProperties": false,
+            "required": ["errorName", "errorOrigin", "errorType"],
+            "properties": {
+                "errorName": {
+                    "type": "string"
```
ege0zcan, line 27: Would it make sense to strongly type error names as well, similar to what you did for errorOrigin below?

ege0zcan (approved):

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -79,6 +86,18 @@ export class OperationalTelemetryService implements OperationalTelemetry {
         if (!this.telemetryOptOut) {
             this.startupSdk()
         }
+
+        // Registering process events callbacks once
+        process.on('uncaughtException', async () => {
+            // Metrics and spans are force flushed to their exporters on shutdown.
+            await this.shutdownSdk()
+            process.exit(1)
```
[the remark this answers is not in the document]
natdurlik: I added it here so we can `await` for `this.shutdownSdk()`. Unfortunately we can only have one on `uncaughtException` handler, but we can expose `shutdownSdk()` method and await for it in standalone.ts. Let's leave this change for the future, if needed.

PR #393, opened 2025-03-28
feat: add agent interface for tool creation and usage

kmile:
## Problem
AI assistants often provide tool usage to interact with the local environment. This enhances the interactivity and effectiveness of the interaction.

## Solution
Provide an interface to keep track of, validate, and run tools.

## License


runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+     * called using `runTool` to ensure tool name lookup and schema validation happen.
+     *
+     * @param spec Tool Specification
+     * @param handler The async method to execute when the tool is called
+     * @returns A tool object for local reference.
+     */
+    tool: <T extends InferSchema<S['inputSchema']>, S extends ToolSpec, R>(
+        spec: S,
+        handler: (input: T) => Promise<R>
+    ) => Tool<T, R>
```
viktorsaws, line 128: Why we return Tool here? If we want to prevent using result of this method, should we not return Tool object, but some flag instead?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+     * Add a tool to the local tool repository. Tools with the same name will be overwritten.
+     *
+     * Instead of using the result of this method, for normal tool execution, the tool should be
+     * called using `runTool` to ensure tool name lookup and schema validation happen.
+     *
+     * @param spec Tool Specification
+     * @param handler The async method to execute when the tool is called
+     * @returns A tool object for local reference.
+     */
+    tool: <T extends InferSchema<S['inputSchema']>, S extends ToolSpec, R>(
```
viktorsaws, line 125: Would renaming this method to something like `registerTool` be clearer?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+}
+
+export type Tools = ToolSpec[]
+export type BedrockTools = {
+    toolSpecification: Omit<ToolSpec, 'inputSchema'> & { inputSchema: { json: ToolSpec['inputSchema'] } }
+}[]
+
+export type Agent = {
+    /**
+     * Add a tool to the local tool repository. Tools with the same name will be overwritten.
```
viktorsaws, line 116: Shall it throw an error instead of overriding to prevent conflicts?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Get the list of tools in the local tool repository.
+     * @param options Options for the format of the output. Can be either 'bedrock' or 'mcp' (the default)
+     * @returns The tool repository in the requested output format
+     */
+    getTools: <T extends GetToolsOptions>(options?: T) => T extends { format: 'bedrock' } ? BedrockTools : Tools
+}
+
+export const newAgent = (): Agent => {
```
viktorsaws, line 150: We shouldn't have implementation in `server-interface` folder, it's better to move it to `runtimes/agent`

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+                validate: (input: InferSchema<S['inputSchema']>) => {
+                    return validator(input)
+                },
+                invoke: handler,
+                name: spec.name,
+                description: spec.description,
+                inputSchema: spec.inputSchema,
+            }
+
+            tools[spec.name] = tool
```
viktorsaws, line 170: Can we add logging to all API calls here? Useful for debugging, but can be a follow-up

runtimes/README.md
```diff
@@ -272,6 +272,120 @@ Consider using LSP [ShowMessage notification](https://microsoft.github.io/langua
...
 
+### Agent Tools
+In order to create agentic applications Servers can define tools that can be sent to a model backend for use. Tools can be created in multiple servers, and are shared between all servers in the same runtime. There is no JSON/RPC component to Tool usage, Tools are all maintained in Runtime memory.
+
+Tool invocations can use LSP, Workspace or Chat capabilities to interact with the client environment.
+
+The agent feature uses the following interface:
+
+```ts
+export type Agent = {
```
viktorsaws, line 283: Is `Agent` a good name for this feature? Looking at what this class does, it's more like tools/agents manager than Agent itself.

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+            const tool = tools[toolName]
+            if (!tool) {
+                throw new Error(`Tool ${toolName} not found`)
+            }
+
+            if (!tool.validate(input)) {
+                throw new Error(`Input for tool ${toolName} is invalid`)
+            }
+
+            return tool.invoke(input)
```
viktorsaws, line 185: Should this call be `await`-ed? 
Also maybe safer to wrap it in try-catch - in case caller server does not handle exceptions, this call can result in unhandledRejection with unexpected side-effects.

runtimes/README.md
```diff
@@ -272,6 +272,120 @@ Consider using LSP [ShowMessage notification](https://microsoft.github.io/langua
...
+     * Add a tool to the local tool repository. Tools with the same name will be overwritten.
+     * 
+     * Instead of using the result of this method, for normal tool execution, the tool should be
+     * called using `runTool` to ensure tool name lookup and schema validation happen.
+     * 
+     * @param spec Tool Specification
+     * @param handler The async method to execute when the tool is called
+     * @returns A tool object for local reference.
+     */
+    tool: <T extends InferSchema<S['inputSchema']>, S extends ToolSpec, R>(spec: S, handler: (input: T) => Promise<R>) => Tool<T, R>
```
jpinkney-aws, line 294: should this be something like `addTool`/`createTool`/`register`? Looking at:
```
AGENT.tool(SOME_TOOL_SPEC, SOME_TOOL_HANDLER)
```
from the test unless you specifically know what `.tool` is doing it's a bit hard to read that code and understand what that lines doing

runtimes/server-interface/agent.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    it('should throw if the tool input does not validate', async () => {
+        AGENT.tool(SOME_TOOL_SPEC, SOME_TOOL_HANDLER)
+        assert.rejects(async () => {
+            await AGENT.runTool(SOME_TOOL_SPEC.name, { test: 1 })
+        }, Error)
+    })
+
+    it('should execute the one correct tool if multiple are available', async () => {
```
jpinkney-aws, line 53: what does "correct tool" mean here? Maybe:
```suggestion
    it('should execute the first registered tool if multiple are available', async () => {
```
?

runtimes/server-interface/agent.ts
kmile: Yeah, I wondered if there were use cases where you want to manually invoke them. It's a good point, might delete it.

runtimes/README.md
kmile: My thinking is an Agent is typically a collection of tools invoked based on model output, which might have different combinations / goals / intents. So I expected the Tools part to be extended with more functionality, and I don't want to make it too narrow and then need a breaking change later.

runtimes/server-interface/agent.ts
kmile: Exceptions (well, `Promise.reject`) is already part of the API, so I don't know how to model that except introducing something like a `Maybe<T>` or a `Try<T>` or an `Either<T>` (which is effectively what `Promise<T>` does as well, but that allows `await` to turn it into a runtime exception again.)

runtimes/README.md
kmile: @viktorsaws had the same feedback. My thinking was based on the method signature and the docs, that this would be clear. But I'll rename it!

mr-lee: +1 to addTool

runtimes/server-interface/agent.test.ts
mr-lee: generally the LLM's will decide which tool is correct.  Is that what's being tested here?

runtimes/server-interface/agent.ts
mr-lee: I don't think we want this - a ton of tools will share names, especially given what I'm seeing.  I think we will want to namespace them to the servers.  Thoughts?

runtimes/README.md
mr-lee: or registerTool (just not tool alone)

runtimes/README.md
```diff
@@ -272,6 +272,120 @@ Consider using LSP [ShowMessage notification](https://microsoft.github.io/langua
...
+A typical agent loop would look like this:
+
+```ts
+const AgentServer: Server = ({ chat, agent }) => {
+
+    agent.tool({
+        name: 'use_greeting',
+        description: 'A kind greeting to the user'
+        inputSchema: {
+            type: 'object',
```
justinmk3, line 326: off-topic: if we are validating schemas at runtime we should consider https://github.com/colinhacks/zod

runtimes/server-interface/agent.ts
volodkevych: Who will be registering tools - server only? Will there be any runtime-registered tools? Any strong reason to put this  tools lib into runtimes?

volodkevych: Where is requirement to have tool names unique coming from? Aren't LLMs deciding what tool to use based on description?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+        },
+
+        runTool: async (toolName: string, input: any) => {
+            const tool = tools[toolName]
+            if (!tool) {
+                throw new Error(`Tool ${toolName} not found`)
+            }
+
+            if (!tool.validate(input)) {
+                throw new Error(`Input for tool ${toolName} is invalid`)
```
volodkevych, line 182: Is it possible to provide some details of what is invalid?

runtimes/server-interface/agent.ts
```diff
@@ -0,0 +1,215 @@
...
+export type Tool<T, R> = {
+    name: string
+    description: string
+    inputSchema: ObjectSchema
+    validate: (input: T) => boolean
+    invoke: (input: T) => Promise<R>
+}
+
+export type GetToolsOptions = {
+    format: 'bedrock' | 'mcp'
```
volodkevych, line 106: Why do we need to differentiate them?

runtimes/server-interface/agent.test.ts
kmile: No, this is just picking the right one by name :) updated the description.

runtimes/server-interface/agent.ts
mr-lee: pretty much all of the tools initially would be runtime-registered, I think.

the primary use of MCP today seems client-local, similar to what LSP has done.

kmile: For that reason I left it as is: no conflicts are expected when the tools are registered inside a runtime, since the goal of putting them inside `Server` is to make them reusable and composable. There should be no need to have 3 different `fs-write` tools in one build.

Once we get to external tools, like through MCP, those should definitely be namespaced. And if we do run into conflicts that are necessary, we can either namespace then or create a way to make either namespaces or sub-views on the `ToolSpec` to make sure one `Server` can use one tool, and another `Server` can use another.

runtimes/server-interface/agent.ts
kmile: This is meant as a short-hand for the model API. Right now MCP has one shape, but Bedrock has a slightly different shape. The tools are stored in MCP format, but given that we're integrating with an API using the Bedrock spec I thought it made sense to give a type-safe shorthand to convert from one to the other.

viktorsaws (approved):

PR #403, opened 2025-04-01
fix: add requestId to InlineChat response

pras0131:
The PR adds requestId to the response of inline chat, as it's needed by client for emitting telemetry. Similar to how these parameters are passed from InlineCompletion response.

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -115,6 +115,9 @@ export interface ChatMessage {
 // Response for chat prompt request can be empty,
 // if server chooses to handle the request and push updates asynchronously.
 export interface ChatResult extends ChatMessage {}
+export interface InlineChatResult extends ChatMessage {
+    requestId?: string
+}
 
 export type EndChatParams = { tabId: string }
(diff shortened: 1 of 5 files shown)


types/chat.ts
```diff
@@ -115,6 +115,9 @@ export interface ChatMessage {
 // Response for chat prompt request can be empty,
 // if server chooses to handle the request and push updates asynchronously.
 export interface ChatResult extends ChatMessage {}
+export interface InlineChatResult extends ChatMessage {
+    requestId?: string
```
ege0zcan, line 119: what is requestId needed for?

nborges-aws: RequestId is needed for telemetry purposes. Since language server won't keep session data for inline chat, requestId will be taken from the chat response and included when building the inlineChatSessionResult object in the client. Otherwise there would be no way to link the requestId to the sessionResult object, since tabId doesn't exist for inline.

francescoopiccoli (approved):

PR #408, opened 2025-04-03
feat: add LSP extension for sending telemetry for inline chat result action

pras0131:
## Problem
Add LSP extension to report the result of action taken by user on inline chat response to the LSP server. As of now, the implementation will be limited to logging telemetry for the action taken by the user on the response provided - ACCEPT/REJECT/DISMISS but in future, it will be extended to have trackers (CodeDiffTracker, CodePercentageTracker) similar to what we have for inline completion result: https://github.com/aws/language-servers/blob/ecd755576df2d2e3ca4591cf44b605d7b72b0bc3/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L603 

## License


types/chat.ts
```diff
@@ -20,6 +20,7 @@ export const FILE_CLICK_NOTIFICATION_METHOD = 'aws/chat/fileClick'
 export const INLINE_CHAT_REQUEST_METHOD = 'aws/chat/sendInlineChatPrompt'
 export const CONTEXT_COMMAND_NOTIFICATION_METHOD = 'aws/chat/sendContextCommands'
 export const CREATE_PROMPT_NOTIFICATION_METHOD = 'aws/chat/createPrompt'
+export const LOG_INLINE_CHAT_RESULT_NOTIFICATION_METHOD = 'aws/chat/logInlineChatResult'
```
justinmk3, line 23: Why do we need to add a method to the protocol just to log telemetry? 

This should be a basic feature of every lsp sever response. We shouldn't need explicit methods just to log telemetry results.

pras0131: We do follow the similar pattern for inline completion result action as well [here](https://github.com/aws/language-servers/blob/996a422665f95656a481a766c8facfd7636ba2ba/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L517)

This is to move telemetry emission to Flare, so clients have a single place to emit telemetry.

justinmk3: My question applies to that case too. Lots of questionable things exist, doesn't mean we should continue them. And it's not too late to change this pattern, though it will be if we keep repeating it.

pras0131: We wish to emit the telemetry to CWSPR DW as we do for other use cases through servers itself, that's why it's there. Also, there has to be some way for server to know what was accepted/rejected etc and this is one way. If not this, what and how do you suggest otherwise?

justinmk3: Suggestion from my first comment:

> This should be a basic feature of every lsp sever response. We shouldn't need explicit methods just to log telemetry results.

Alternative suggestion:

Have *one* single generic "telemetry" notification. Why model specific events for telemetry? The client action is always the same: log the telemetry (and map various fields, which should be automated, not manually-mapped for every telemetry emission from the LSP server).

types/chat.ts
```diff
@@ -303,8 +304,29 @@ export interface CreatePromptParams {
...
 }
 
-// history
+export interface ProgrammingLanguage {
+    languageName: string
+}
+
+export type InlineChatUserDecision = 'ACCEPT' | 'REJECT' | 'DISMISS' | string
 
+export interface InlineChatResultParams {
```
viktorsaws, line 313: @pras0131 please document every new Protocol addition and meaning of all fields in doc strings. It should be very clear what is expected from client to pass here, and we should follow LSP spec example. You can see example of documentation I did in latest PR https://github.com/aws/language-server-runtimes/pull/433/files

types/chat.ts
viktorsaws: If this is for similar use case as [`onLogInlineCompletionSessionResultsHandler`](https://github.com/aws/language-servers/blob/996a422665f95656a481a766c8facfd7636ba2ba/server/aws-lsp-codewhisperer/src/language-server/inline-completion/codeWhispererServer.ts#L517), then it needs to be implemented the proposed way. 

Interaction flow with InlineCompletion/InlineChat goes like this: 
1. Client sends request to server to get inline suggestion (`aws/chat/sendInlineChatPrompt`)
2. Server returns suggestions
3. Client shows suggestions in UI
4. User interacts with suggestions (accepts, rejects, cancels, etc.)
5. Client applies user choice 
6. Client sends results and data about interaction back to server (`aws/chat/logInlineChatResult`)

This protocol is not for server sending info to client, but opposite - client sends a result of interaction with chat suggestions back to the server. With `onLogInlineCompletionSessionResultsHandler`, we need this to be able to calculate telemetry correctly on the server, for which it needs some data from client-side interaction, which is not available otherwise. Reason we did it this way for InlineCompletions was to encapsulate business logic of computing telemetry events in Language Server.


@pras0131 we need to document the purpose of each new API very clearly in code `docstring`, similar to LSP specs, otherwise all these changes are very confusing to reason about.

types/chat.ts
```diff
@@ -303,8 +304,29 @@ export interface CreatePromptParams {
...
+    requestId: string
+    timestamp: Date
+    inputLength?: number
+    numSelectedLines?: number
+    numSuggestionAddChars?: number
+    numSuggestionAddLines?: number
+    numSuggestionDelChars?: number
```
viktorsaws, line 320: It's better to use full words in protocol, it's clearer and more understandable.

types/chat.ts
```diff
@@ -303,8 +304,29 @@ export interface CreatePromptParams {
...
-// history
+export interface ProgrammingLanguage {
+    languageName: string
+}
+
+export type InlineChatUserDecision = 'ACCEPT' | 'REJECT' | 'DISMISS' | string
 
+export interface InlineChatResultParams {
+    requestId: string
+    timestamp: Date
```
viktorsaws, line 315: What is this timestamp needed for and why it is required?

pras0131: Same as above, required by InlineChatEvent but I do agree that we can just simply calculate it at server end instead of client passing it.

types/chat.ts
pras0131: I can add a doc string to inlineChatResultNotificationType but adding a doc string to all field does not seem to be required here post the renaming of variables to fuller names as they are clearly indicate their purpose.

types/chat.ts
```diff
@@ -311,8 +312,28 @@ export interface CreatePromptParams {
...
+export type InlineChatUserDecision = 'ACCEPT' | 'REJECT' | 'DISMISS' | string
 
+export interface InlineChatResultParams {
+    requestId: string
+    inputLength?: number
+    numberOfSelectedLines?: number
+    numberOfCharactersAddedFromSuggestion?: number
+    numberOfLinesAddedFromSuggestion?: number
+    numberOfCharactersDeletedFromSuggestion?: number
+    numberOfLinesDeletedFromSuggestion?: number
```
justinmk3, line 328: this is a new convention that doesn't follow any existing convention in this codebase. prefixing "numberOf" in front of a number just isn't discoverable or helpful.

what is discoverable, is using a common prefix that groups things in a meaningful way. here, the grouping is "suggestion".

and, the type already mentions `number`, so there's no need to repeat that in the name.


```suggestion
    selectedLines?: number
    suggestionAddedChars?: number
    suggestionAddedLines?: number
    suggestionDeletedChars?: number
    suggestionDeletedLines?: number
```

viktorsaws (approved):

PR #410, opened 2025-04-04
feat: added chat history protocol

volodkevych:
## Problem

Conversation history support needs to be added to chat protocol and server interface.

## Solution

Added the following new methods to chat protocol and server interface:
- `aws/chat/listConversations` (supports also filtering)
- `aws/chat/conversationClick` (will open conversation on item click, or will execute action - delete or export - if item-related action is clicked)

## License


anasstahr (approved):

runtimes/README.md
```diff
@@ -186,6 +186,9 @@ The runtime supports chat by default
...
 | Send create prompt event that triggers new prompt creation flow on server. | `aws/chat/createPrompt`          | `CreatePromptParams`          | [Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage) Client to Server | n/a             |
+| Send request to list the conversations available in history: all or based on filter if provided. As there can be several filter opitons used, the filter in the request is a map of filter option ids to corresponding values. Possible filter options are expected to be provided in the previous `listConversations` result before filter can be used. | `aws/chat/listConversations`          | `ListConversationsParams`          | [Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) Client to Server | `ListConversationsResult`             |
```
kmile, line 189: nit: typo on "opitons".

runtimes/README.md
```diff
@@ -186,6 +186,9 @@ The runtime supports chat by default
...
+| Send conversation or conversation action click event. | `aws/chat/conversationClick`          | `ConversationClickParams`          | [Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage) Client to Server | n/a             |
```
kmile, line 190: This seems really UI focused (`click`). What is the user trying to achieve with the click?

runtimes/protocol/chat.ts
```diff
@@ -46,6 +46,13 @@ import {
     ContextCommandParams,
     CREATE_PROMPT_NOTIFICATION_METHOD,
     CreatePromptParams,
+    ListConversationsParams,
+    ListConversationsResult,
+    LIST_CONVERSATIONS_REQUEST_METHOD,
+    ConversationClickParams,
+    ConversationsUpdateParams,
+    CONVERSATION_CLICK_NOTIFICATION_METHOD,
+    CONVERSATIONS_UPDATE_NOTIFICATION_METHOD,
```
kmile, line 55: I'm always curious why all of these types come from `./lsp`.

runtimes/README.md
volodkevych: User would open a conversation if he clicks on a conversation or execute action if they click on an action within that conversation. There are several ways to model it, but we already use this 'click' event approach in several other cases.

runtimes/README.md
```diff
@@ -186,6 +186,9 @@ The runtime supports chat by default
...
+| Send event with updated conversations. Can be useful to update conversations list after an item was deleted from it, if client does not handle deletion automatically or in any other case when business logic needs to update conversations without explicit customer request. | `aws/chat/sendConversationsUpdate`          | `ConversationsUpdateParams`          | [Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage) Server to Client | n/a             |
```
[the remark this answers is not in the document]
volodkevych: This is about list, not message inside conversation. At this point of time I need it to update the list after delete: given delete is a notification from client to server (`aws/chat/conversationClick`), I cannot 'pull' list from client to server in the same place, because it will cause race condition. The other way to approach it would be to do request not notification for delete, then new list could be pulled as a next step, but then I would also need to split `aws/chat/conversationClick` to 3 separate events I think.

runtimes/protocol/chat.ts
volodkevych: Somebody decided to re-export all the types from `lsp.ts` at some point of time.

runtimes/README.md
ege0zcan: what do you think about `chat/conversationAction` or `chat/conversationCommand`?

runtimes/protocol/chat.ts
```diff
@@ -105,9 +112,26 @@ export const chatUpdateNotificationType = new ProtocolNotificationType<ChatUpdat
...
+// context
 export const contextCommandsNotificationType = new ProtocolNotificationType<ContextCommandParams, void>(
     CONTEXT_COMMAND_NOTIFICATION_METHOD
 )
 export const createPromptNotificationType = new ProtocolNotificationType<CreatePromptParams, void>(
     CREATE_PROMPT_NOTIFICATION_METHOD
 )
+
+// history
+export const listConversationsRequestType = new ProtocolRequestType<
```
ege0zcan, line 125: I think it is better to use `AutoParameterStructuresProtocolRequestType` here, having both `ProtocolRequestType` and `AutoParameterStructuresProtocolRequestType` in the codebase can be confusing as one might wonder if the some requests were intentionally done one way while some were done the other way

And as for why we have AutoParameterStructuresProtocolRequestType, as far as I remember, using this type prevented some unwanted side effects or stuff breaking when we used `npm link` across language-servers and runtimes packages for example

@kmile was recently facing some issues so he might have fresher context

volodkevych: ok, changed

runtimes/README.md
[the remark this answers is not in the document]
volodkevych: This event also opens the conversation if no action was clicked, but the conversation itself.

types/chat.ts
```diff
@@ -284,3 +290,51 @@ export interface ContextCommandParams {
...
+    filterOptions?: FilterOption[]
+    list: ConversationItemGroup[]
+}
+
+export interface ListConversationsResult extends ConversationsList {}
+export interface ConversationsUpdateParams extends ConversationsList {}
+
+export type ConversationAction = 'delete' | 'export'
+
+export interface ConversationClickParams {
```
ege0zcan, line 337: Why is `action` optional? What does it mean when a ConversationClick happens without an action?

types/chat.ts
```diff
@@ -284,3 +290,51 @@ export interface ContextCommandParams {
...
+    items?: ConversationItem[]
+}
+
+export interface ListConversationsParams {
+    // key maps to id in FilterOption and value to corresponding filter value
+    filter?: Record<string, FilterValue>
+}
+
+export interface ConversationsList {
+    hearer?: { title: string }
```
ege0zcan, line 327: typo: header

types/chat.ts
```diff
@@ -284,3 +290,51 @@ export interface ContextCommandParams {
...
+export interface Action {
+    id: string
+    icon?: IconType
+    text: string
+}
+export interface ConversationItem {
+    id: string
+    icon?: IconType
+    description?: string
+    actions?: Action
```
ege0zcan, line 312: `actions` suggests there can be multiple actions here. We should either rename to `action` or make the type `Action[]`

ege0zcan (approved):

types/chat.ts
volodkevych: It means to open it - as in all the other `...Click` events.

types/chat.ts
volodkevych: fixed

types/chat.ts
volodkevych: Will state it in the readme explicitly.

ege0zcan: Ah ok, good to know thank you

runtimes/README.md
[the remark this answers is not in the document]
volodkevych: As discussed, changed `aws/chat/conversationClick` from notification to request. It will allow us to request `aws/chat/listConversations` to update the items after successful action execution without the need of additional method `aws/chat/sendConversationsUpdate` and without risk of race condition.

ege0zcan (approved):

types/chat.ts
```diff
@@ -134,7 +136,8 @@ export interface QuickActionCommand {
     icon?: IconType
 }
 
-export type IconType = 'file' | 'folder' | 'code-block' | 'list-add' | 'magic' | 'help' | 'trash'
+export type ContextCommandIconType = 'file' | 'folder' | 'code-block' | 'list-add' | 'magic'
+export type IconType = ContextCommandIconType | 'help' | 'trash' | 'search' | 'calendar' | string
```
ege0zcan, line 140: Why do we allow other `string` values here?

volodkevych: The list of this icons is growing fast - I'm not sure we want to list all of the explicitly. I thought it makes sense to have most used ones explicitly listed, however we should not update the protocol for every new icon added.

PR #411, opened 2025-04-07
feat: extend tabdata interface to include support for welcome screen

anasstahr:
## Problem
We don't have a way to integrate the current welcome screen in Q through flare

## Solution
Extending to protocol so we can integrate the welcome screen UI in flare

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -158,18 +158,21 @@ export interface QuickActions {
 }
 
 export interface TabData {
-    title: string
-    placeholderText?: string
-    placeHolderLabel?: string
+    tabHeader?: TabHeader
+    promptInput?: PromptInput
     compactMode?: boolean
     messages: ChatMessage[]
-    tabHeaderDetails?: TabHeaderDetails
 }
 
-export interface TabHeaderDetails {
+export interface TabHeader {
     icon?: IconType
     title?: string
     description?: string
+}
+
+export interface PromptInput {
+    placeHolderText?: string
+    label?: string
 }
 
 /**
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -159,6 +159,7 @@ export interface QuickActions {
 }
 
 export interface TabData {
+    title?: string
     tabHeader?: TabHeader
     promptInput?: PromptInput
     compactMode?: boolean
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -28,6 +28,7 @@ export interface ChatItemAction {
     description?: string
     type?: string
     status?: string
+    id?: string
 }
 
 export interface SourceLink {
@@ -171,7 +172,7 @@ export interface TabHeader {
 }
 
 export interface PromptInput {
-    placeHolderText?: string
+    placeholderText?: string
     label?: string
 }
 
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -27,6 +27,7 @@ export interface ChatItemAction {
     disabled?: boolean
     description?: string
     type?: string
+    status?: string
 }
 
 export interface SourceLink {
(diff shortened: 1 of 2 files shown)


types/chat.ts
```diff
@@ -155,7 +158,10 @@ export interface QuickActions {
 }
 
 export interface TabData {
```
volodkevych, line 160: Should we add new properties to support setting tab header? (`tabHeaderDetails` property on vscode repo)

types/chat.ts
```diff
@@ -155,7 +158,10 @@ export interface QuickActions {
...
+    title: string
     placeholderText?: string
+    placeHolderLabel?: string
+    centerContent?: boolean
```
volodkevych, line 164: Is it for `compactMode`?

types/chat.ts
volodkevych: I see you added `title`, but we still need `description` and header-specific `icon`. Shall we introduce this `tabHeader` as a new type in the protocol?

types/chat.ts
anasstahr: Yes, do we want to change the name to `compactMode` ?

volodkevych: `compactMode` is sounds better for what that flag is doing

types/chat.ts
```diff
@@ -155,8 +158,18 @@ export interface QuickActions {
 }
 
 export interface TabData {
+    title: string
```
volodkevych, line 161: Now, `title` is not needed here, right?

types/chat.ts
```diff
@@ -155,8 +158,18 @@ export interface QuickActions {
...
     placeholderText?: string
+    placeHolderLabel?: string
```
volodkevych, line 163: nit: I did not know about `placeHolderLabel` when I added `placeholderText`. Maybe we could group those 2 props under `Placeholder` type. I know it's technically a breaking change, but nobody uses `TabData` yet, it was added recently, so I think we can still make this change without breaking any dependency.

types/chat.ts
anasstahr: you're right, missed that

types/chat.ts
anasstahr: Sure that sounds good. Does the following sound ok?

```
placeHolder: PlaceHolder

export interface PlaceHolder{
   text?: string
   label?: string
}
```

volodkevych: nit: `placeholder` should be 1 word, no camelCase?

volodkevych (approved):

volodkevych (approved):

anasstahr:
I changed the following after going through mynahUI again:
- Adding a signal for the welcome Counter that the vscode toolkit also uses to know when to update an internal state for the welcomeScreen
- Added the title back in `tabData`. This is different from the `tabHeader.title`. `tabData.title` is used for the actual title of the tab while `tabHeader.title` is used for text within the tab

volodkevych (approved):

types/chat.ts
```diff
@@ -155,10 +159,24 @@ export interface QuickActions {
...
     messages: ChatMessage[]
 }
 
+export interface TabHeader {
+    icon?: IconType
+    title?: string
+    description?: string
+}
+
+export interface PromptInput {
```
justinmk3, line 175: Why don't these have docstrings that give insight into how they should be used? Is that documented somewhere else? If not, can we make that a habit for all runtimes PRs?

types/chat.ts
```diff
@@ -100,6 +102,8 @@ export interface FileList {
 export interface ChatMessage {
     type?: 'answer' | 'prompt' | 'system-prompt' // will default to 'answer'
     body?: string
+    buttons: ChatItemAction[]
```
viktorsaws, line 105: Why `buttons` are required field? It breaks compatibility in language-server repository, and code either should be adjusted there or this should be optional.

volodkevych: FYI: this PR was reverted

PR #412, opened 2025-04-07
fix(log): remove logging class names

francescoopiccoli:
## Problem
Since truncated logging from bundling minimization modifies and makes class/function names unrecognizable, we remove class name logging from runtimes, instead and we log directly hardoded class names from language-servers (see https://github.com/aws/language-servers/pull/909)

## License


justinmk3 (approved):
This is a "fix", not "chore".

PR #415, opened 2025-04-08
fix: update Dirent interface. 

Hweinstock:
## Problem
This Dirent interface is difficult to work with for two reasons:
- it uses deprecated and confusing path field ([https://github.com/nodejs/node/issues/51955#issuecomment-1977131319 ](https://github.com/nodejs/node/issues/51955#issuecomment-1977131319)). To summarize, path is the same as parentPath on node 20, but not node 18. this is not clear in the name. 
- does not provide a way to check for links, which are returned by fs.readdir.

## Solution
- update interface to avoid deprecated field, and expose the predicate for checking links. 

## License

### DIFF
--- a/runtimes/server-interface/workspace.ts
+++ b/runtimes/server-interface/workspace.ts
@@ -5,7 +5,7 @@ interface Dirent {
     isFile(): boolean
     isDirectory(): boolean
     name: string
-    path: string
+    parentPath: string
 }
 
 /**
--- a/runtimes/server-interface/workspace.ts
+++ b/runtimes/server-interface/workspace.ts
@@ -4,6 +4,7 @@ import { TextDocument, WorkspaceFolder } from '../protocol'
 interface Dirent {
     isFile(): boolean
     isDirectory(): boolean
+    isSymbolicLink(): boolean
     name: string
     parentPath: string
 }


runtimes/server-interface/workspace.ts
```diff
@@ -4,8 +4,9 @@ import { TextDocument, WorkspaceFolder } from '../protocol'
 interface Dirent {
     isFile(): boolean
     isDirectory(): boolean
+    isSymbolicLink(): boolean
     name: string
-    path: string
```
kmile, line 8: Does this break anyone in https://github.com/aws/language-servers/tree/main/server ?

Hweinstock: Good callout. I searched for references and didn't see any (besides the ones I add in https://github.com/aws/language-servers/pull/923). Those will have to migrated on the version bump, but allows us to remove those workarounds. 

This type is implicitly used here: https://github.com/aws/language-servers/blob/c1aaec06b70f4ef9d5e2a7ad0d1cc4d5d6955087/server/aws-lsp-codewhisperer/src/language-server/securityScan/dependencyGraph/dependencyGraph.ts#L110, but this change shouldn't affect its use case.

kmile: Ack, thanks for following up!

PR #421, opened 2025-04-09
feat(runtimes): introduce shared registry feature

patrickvbeurden:
## Problem

Runtimes can consist of multiple Servers, and use cases have popped up where these servers end up having shared dependencies. However, the Servers cannot communicate, which leads to issues such as duplication or dependencies opting to hijack LSP features from the server.

## Solution

Introduce a new feature that allows servers in the same runtime to expose shared functionality amongst each other. This `SharedRegistry` feature allows to register named handlers and after that any server in the runtime can invoke this handler by name. The implementation is very similar to adding and invoking `Tools` in the recently added `Agent` feature ([#393](https://github.com/aws/language-server-runtimes/pull/393)). Since this feature is intended for inter-server communication, we can opt to simplify the feature interface and be less strict with validation.

Proposed interface:

```ts
/**
 * The SharedRegistry feature allows a Server to vend shared functionality to other
 * Servers in the same runtime by registering handlers which can then be invoked by
 * name in the other Servers.
 */
export type SharedRegistry = {
    /**
     * Registers a handler to the shared registry, after which it can be invoked
     * through the `invokeHandler` method.
     *
     * @param name The name under which the handler should be registered
     * @param handler The async function to execute when the handler is invoked by name
     */
    registerHandler: <T, R>(name: string, handler: (input: T) => Promise<R>) => void

    /**
     * Invokes the requested handler in the shared registry.
     *
     * This function throws if the handler is not found, and **does not**
     * handle any errors thrown by the invoked handler.
     *
     * @param handlerName The handler in the shared registry that should be invoked
     * @param input The input to the handler (if any)
     * @returns A promise for the result of the handler invocation
     */
    invokeHandler: (handlerName: string, input?: any) => Promise<any>

    /**
     *
     * @returns A list of names of the currently registered handlers
     */
    listHandlers: () => string[]
}
```

Usage example from README:

```ts
interface CommonFunctionality {
    getSomething(): Promise<Something>
    doSomething(something: Something): Promise<Something>
}

const CommonFunctionalityServer: Server = ({ sharedRegistry }) => {
    const commonFunctionality: CommonFunctionality = {
        // ...
    }

    sharedRegistry.registerHandler('getSomething', () => commonFunctionality.getSomething()))
    sharedRegistry.registerHandler('doSomething', (something: Something) => commonFunctionality.doSomething(something)))
}


const OtherServer: Server = ({ lsp, sharedRegistry } => {
    let something: Something

    lsp.onCompletion(() => {
        try {
            something = sharedRegistry.invokeHandler('getSomething')
            // ...
            something = sharedRegistry.invokeHandler('doSomething', something)
            // ...
        } catch(error) {
            // ...
        }
    })
})
```

Testing: added unit tests, tested with `language-servers` that different servers can indeed use each other registered handlers.

## License


runtimes/runtimes/shared-registry.ts
```diff
@@ -0,0 +1,33 @@
+import { SharedRegistry } from '../server-interface/shared-registry'
+
+type SharedHandler<T, R> = {
+    name: string
+    invoke: (input: T) => Promise<R>
+}
+
+export const newSharedRegistry = (): SharedRegistry => {
```
viktorsaws, line 8: nit: better naming for factory function would be `createSharedRegistry`

runtimes/runtimes/standalone.test.ts
```diff
@@ -126,12 +127,12 @@ describe('standalone', () => {
     describe('features', () => {
         let features: Features
 
-        beforeEach(() => {
-            standalone(props)
-            features = stubServer.getCall(0).args[0]
-        })
-
         describe('Workspace', () => {
+            beforeEach(() => {
```
viktorsaws, line 131: nit, can be follow-up fix: Copy-pasting beforeEach block is not intuitive setup. If you want to avoid usage of `generateSharedRegistryFeatureTest` conflicting with setup for other tests, you can invoke `generateSharedRegistryFeatureTest(standalone)` outside of shared `describe('features', () => {}` block.

viktorsaws (dismissed):

runtimes/runtimes/shared-registry.ts
```diff
@@ -0,0 +1,33 @@
...
+    const sharedHandlers: Record<string, SharedHandler<any, any>> = {}
+
+    return {
+        registerHandler: <T, R>(name: string, handler: (input: T) => Promise<R>) => {
```
justinmk3, line 12: How is this different than an `EventEmitter`? Should we use that interface instead of creating a bespoke one?

runtimes/server-interface/shared-registry.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * The SharedRegistry feature allows a Server to vend shared functionality to other
+ * Servers in the same runtime by registering handlers which can then be invoked by
+ * name in the other Servers.
+ */
+export type SharedRegistry = {
```
justinmk3, line 6:
```suggestion
export type EventBus = {
```

PR #424, opened 2025-04-10
feat: inject home directory into runtimes. 

Hweinstock:
## Problem
The home directory is currently fetched via https://github.com/aws/language-servers/blob/ecd755576df2d2e3ca4591cf44b605d7b72b0bc3/core/aws-lsp-core/src/util/path.ts#L96 for components that require it, but this is not available on web. 

Rather than adding web-safe checks around this, we should inject this into the runtime via the `Workspace` interface. 

Injecting this behavior will also make it easier to test certain scenarios as we can directly manipulate the home directory without mocking. 

## Solution
- Add `workspace.fs.getUserHomeDir` that resolves to `os.homedir()` on standalone processes. 
- Default to an empty string in web browsers where there is no concept of home directories. 
- For fetching server directory on web, attempt to use `initializationOptions.clientDataFolder` and fallback to server name. Previously this resolved to empty string always. Empty string is now reserved for the home directory. 

## Note
This PR is done with the assumption that home directories will not be used directly unless explicitly required by a feature. For storage purposes, the server storage should the default. [Example](https://github.com/aws/language-servers/blob/594729d9272109aba7386d08ad1b2f68b493ac63/server/aws-lsp-codewhisperer/src/language-server/agenticChat/tools/chatDb/chatDb.ts#L47): could be changed to avoid browser check and simply use server storage. 

## Follow-up:
On version bump, replace references to `os.homedir()` with `workspace.fs.getUserHomeDir`. 

## License

### DIFF
--- a/runtimes/runtimes/base-runtime.ts
+++ b/runtimes/runtimes/base-runtime.ts
@@ -119,7 +119,7 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
             getServerDataDirPath: serverName =>
                 lspRouter.clientInitializeParams?.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
             getTempDirPath: () => '/tmp',
-            getUserHomeDir: () => '',
+            getUserHomeDir: () => '/home/user',
             readFile: (_path, _options?) => Promise.resolve(''),
             readdir: _path => Promise.resolve([]),
             isFile: _path => Promise.resolve(false),
--- a/runtimes/runtimes/base-runtime.ts
+++ b/runtimes/runtimes/base-runtime.ts
@@ -117,7 +117,7 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
             getServerDataDirPath: serverName =>
-                lspRouter.clientInitializeParams!.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
+                lspRouter.clientInitializeParams?.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
             getTempDirPath: () => '/tmp',
             getUserHomeDir: () => '',
             readFile: (_path, _options?) => Promise.resolve(''),
--- a/runtimes/runtimes/base-runtime.ts
+++ b/runtimes/runtimes/base-runtime.ts
@@ -116,8 +116,10 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
             copyFile: (_src, _dest, _options?) => Promise.resolve(),
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
-            getServerDataDirPath: _serverName => '',
+            getServerDataDirPath: serverName =>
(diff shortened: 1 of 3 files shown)


runtimes/runtimes/base-runtime.ts
```diff
@@ -116,8 +116,10 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
             copyFile: (_src, _dest, _options?) => Promise.resolve(),
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
-            getServerDataDirPath: _serverName => '',
+            getServerDataDirPath: serverName =>
+                lspRouter.clientInitializeParams!.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
```
justinmk3, line 120: since this already handles falsy, can avoid "!"

```suggestion
                lspRouter.clientInitializeParams?.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
```

runtimes/runtimes/base-runtime.ts
```diff
@@ -116,8 +116,10 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
...
             getTempDirPath: () => '/tmp',
+            getUserHomeDir: () => '',
```
justinmk3, line 122: if "/tmp" is a valid default on web, maybe "/home/user" is a valid default too :) 

does `runtimes/webworker.ts` need to auto-create those directories somehow? (probably out of scope for this PR, but just curious)

Hweinstock: Thats a good question. What would creating those directories look like in a web worker since there isn't an actual file system? Is there other other initialization necessary?

justinmk3: Not sure about web *workers*, but web (html) has a filesystem api (isolated per browser-tab): 
https://developer.mozilla.org/en-US/docs/Web/API/FileSystem

I would guess that api (or something like it) is used by web apps like http://github.dev/ for all of the files/directories the user interacts with, including in the terminal/shell that runs in vscode-web.

runtimes/runtimes/base-runtime.ts
```diff
@@ -116,8 +116,10 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
             copyFile: (_src, _dest, _options?) => Promise.resolve(),
             exists: _path => Promise.resolve(false),
             getFileSize: _path => Promise.resolve({ size: 0 }),
-            getServerDataDirPath: _serverName => '',
+            getServerDataDirPath: serverName =>
+                lspRouter.clientInitializeParams?.initializationOptions?.aws?.clientDataFolder ?? `/${serverName}`,
```
volodkevych, line 120: for consistency, I would prefer the logic that we use for `getServerDataDirPath` in standalone: join `clientDataFolder` and `serverName`

volodkevych: The idea of this method is that it returns server-specific directory.

Hweinstock: That makes sense, I changed it to join the paths and use the homedir as the default to be consistent with `getServerDataDirPath` in standalone.

volodkevych (approved):

PR #430, opened 2025-04-14
feat: added basic eslint to repo

anasstahr:
## Problem
There currently doesn't exist any linting in this repo

## Solution
Add linting by using VScode's team eslint as a template. Copied over minimal amount from their .eslint without custom rules. 
I'm also adding service changes to fix the linting errors later, however currently this is a draft of just the eslint changes.

## License


.eslintrc.js
```diff
@@ -0,0 +1,114 @@
...
+            {
+                selector: 'variable',
+                modifiers: ['const'],
+                format: ['camelCase', 'PascalCase', 'UPPER_CASE'],
+            },
+        ],
+        // Avoid accidental use of "==" instead of "===".
+        eqeqeq: 'error',
+        '@typescript-eslint/no-namespace': 'error',
+        '@typescript-eslint/no-floating-promises': 'off', // Temporarily disabled
```
volodkevych, line 59: I would prefer this enabled - it allows to catch when people forget to await.

.eslintrc.js
```diff
@@ -0,0 +1,114 @@
...
+        '@typescript-eslint/no-redundant-type-constituents': 'off', // Temporarily disabled
+        '@typescript-eslint/no-unused-vars': 'error',
+        '@typescript-eslint/no-explicit-any': 'off', // Temporarily disabled
```
volodkevych, line 62: Why disabled?

.eslintrc.js
```diff
@@ -0,0 +1,114 @@
...
+        'unicorn/prefer-modern-math-apis': 'error',
+        'unicorn/prefer-native-coercion-functions': 'error',
+        'unicorn/prefer-reflect-apply': 'error',
+        'unicorn/prefer-string-trim-start-end': 'error',
+        'unicorn/prefer-type-error': 'error',
+        // Discourage `.forEach` because it can lead to accidental, incorrect use of async callbacks.
+        'unicorn/no-array-for-each': 'error',
+        'security-node/detect-child-process': 'error',
+
+        // TODO: do we want to enable this?
```
volodkevych, line 100: Can we remove these comments please?

.eslintrc.js
justinmk3: Definitely valuable to enable this, but if the codebase isn't prepared for it, it may trigger a lot of warnings. Worth a try though.

.eslintrc.js
```diff
@@ -0,0 +1,114 @@
...
+        ecmaVersion: 'latest',
+        sourceType: 'module',
+        requireConfigFile: false,
+    },
+    env: {
+        node: true,
+        mocha: true,
+        es2024: true,
+    },
+    plugins: ['@typescript-eslint', '@stylistic', 'unicorn', 'header', 'security-node'],
```
justinmk3, line 14: Does this codebase already have unicorn etc. in its packages.json dependencies list? If not, they would need to be added.

.eslintrc.js
```diff
@@ -0,0 +1,114 @@
...
+                // Allow underscores.
+                leadingUnderscore: 'allowSingleOrDouble',
+                trailingUnderscore: 'allowSingleOrDouble',
+            },
+            // Allow object properties, enums, and methods to have any format
+            {
+                selector: ['objectLiteralProperty', 'classMethod', 'enumMember'],
+                format: null,
+            },
+            // Allow constants to use UPPER_CASE
```
justinmk3, line 49: UPPER_CASE is really over-used and not helpful (in this codebase), but too late to change it in this codebase.

.eslintrc.js
volodkevych: should be pretty simple to enable it in runtimes

.eslintrc.js
volodkevych: I think TODO is for something that has concrete action plan. If we just add it as a comment, it will stay here forever most likely. I do not think adding this header is anything high priority for now.

.eslintrc.js
anasstahr: Removed the rule for now, don't think it really adds any value

.eslintrc.js
anasstahr: Included a more minimal amount of rules without any unicorn rules for now so I only kept the ts-eslint plugin

.eslintrc.js
anasstahr: Removed!

.eslintrc.js
anasstahr: +1 on justin

anasstahr:
Removed a lot of the rules, we can keep on adding but kept a minimal set. Also kept some of them off, since i'm personally not that familiar with some of these errors.

runtimes/runtimes/chat/baseChat.ts
```diff
@@ -124,15 +124,21 @@ export class BaseChat implements Chat {
     }
 
     public sendChatUpdate(params: ChatUpdateParams) {
-        this.connection.sendNotification(chatUpdateNotificationType.method, params)
+        this.connection
```
volodkevych, line 127: It would be better to change this method return `Promise<void>`.

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -170,7 +167,10 @@ ${JSON.stringify({ ...result.capabilities, ...result.awsServerCapabilities })}`
         const workspaceCapabilities = this.clientInitializeParams?.capabilities.workspace
         if (workspaceCapabilities?.didChangeConfiguration?.dynamicRegistration) {
             // Ask the client to notify the server on configuration changes
-            this.lspConnection.client.register(DidChangeConfigurationNotification.type, undefined)
+            this.lspConnection.client
+                .register(DidChangeConfigurationNotification.type, undefined)
+                .then(() => {})
+                .catch(() => {})
```
volodkevych, line 173: Could we log an error here?

runtimes/runtimes/lsp/router/lspServer.ts
```diff
@@ -31,7 +31,7 @@ export class LspServer {
     readonly notification: Notification
 
     private didChangeConfigurationHandler?: NotificationHandler<DidChangeConfigurationParams>
-    private executeCommandHandler?: RequestHandler<ExecuteCommandParams, any | undefined | null, void>
+    private executeCommandHandler?: RequestHandler<ExecuteCommandParams, unknown, void>
```
volodkevych, line 34: Why did we need this change?

types/chat.ts
```diff
@@ -145,7 +145,7 @@ export interface QuickActionCommand {
 }
 
 export type ContextCommandIconType = 'file' | 'folder' | 'code-block' | 'list-add' | 'magic'
-export type IconType = ContextCommandIconType | 'help' | 'trash' | 'search' | 'calendar' | string
+export type IconType = 'ContextCommandIconType' | 'help' | 'trash' | 'search' | 'calendar'
```
volodkevych, line 148: This is a breaking change - I do not think we can make it. Also, 'ContextCommandIconType' is different from the type above adding more icon types to this type?

anasstahr: > This is a breaking change
Yeah I wasn't so sure how to change this typing. Do we just want to keep it as a `string` type and remove the other hardcoded strings or do we just disable the eslint rule for this specific line

> 'ContextCommandIconType' ...
Missed that one, thanks for pointing it out

PR #432, opened 2025-04-14
feat: show dev profile protocol changes

vidkhath:
## Problem
Q chat introduced new feature which displays the developer profile used. The flare protocol does not support this feature. 
image

## Solution
Extended the flare protocol to introduce a new notification event from server to the client informing them of dev profile changes.

We create a new interface `ShowDevProfileParams` which extends `Partial<ChatMessage>` as the server may specify the body and the type of the message. It also accepts a new `profileName` string which is the dev profile name that is associated

## License


runtimes/protocol/chat.ts
```diff
@@ -105,6 +107,9 @@ export const followUpClickNotificationType = new ProtocolNotificationType<Follow
 export const openTabRequestType = new ProtocolRequestType<OpenTabParams, OpenTabResult, never, void, void>(
     OPEN_TAB_REQUEST_METHOD
 )
+export const showDevProfileType = new ProtocolNotificationType<ShowDevProfileParams, void>(
+    SHOW_DEV_PROFILE_NOTIFICATION_METHOD
```
viktorsaws, line 111: I'm not sure if we want to keep this banner text specifically coupled to Developer profiles feature. I think making this banner more generic will be better and give more flexibility in future in needed.

vidkhath: The banner text can be generic as the ShowDevProfileParams extends ChatMessage which accepts body, type etc. I could rename `showDevProfileType` to `showBannerType` instead, if that helps. WDYT?

types/chat.ts
```diff
@@ -17,6 +17,7 @@ export const OPEN_TAB_REQUEST_METHOD = 'aws/chat/openTab'
 export const CHAT_UPDATE_NOTIFICATION_METHOD = 'aws/chat/sendChatUpdate'
 export const FILE_CLICK_NOTIFICATION_METHOD = 'aws/chat/fileClick'
 export const INLINE_CHAT_REQUEST_METHOD = 'aws/chat/sendInlineChatPrompt'
+export const SHOW_DEV_PROFILE_NOTIFICATION_METHOD = 'aws/chat/showBanner'
```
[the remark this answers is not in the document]
vidkhath: good catch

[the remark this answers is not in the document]
volodkevych: Hmm, `showBanner` looks too UI-specific for me. What if UI does not support banners?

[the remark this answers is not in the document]
volodkevych: After looking into the current protocol again, I think the concept of this tab sticky message fits nicely into `TabData` structure. What do you think about supporting it through `openTab` and `sendChatUpdate`? I know `sendChatUpdate` is not supported in `chat-client` yet, but we could implement the support for it - it may anyway be needed for the whole agentic chat UI experience.

types/chat.ts
```diff
@@ -273,6 +274,10 @@ export interface ChatUpdateParams {
     data?: TabData
 }
 
+export interface ShowBannerParams extends Partial<ChatMessage> {
```
volodkevych, line 277: Why does it extend `Partial<ChatMessage>`?

vidkhath: The fields of ChatMessage are optional in the banner context as the only fields required there would be the profile name and rest can be hardcoded. Vscode implementation for reference: https://github.com/aws/aws-toolkit-vscode/blob/f1dded7edcf7d4d95410f31cfb87da208e91fb27/packages/core/src/amazonq/webview/ui/tabs/generator.ts#L60C15-L68

types/chat.ts
```diff
@@ -18,6 +18,7 @@ export const CHAT_UPDATE_NOTIFICATION_METHOD = 'aws/chat/sendChatUpdate'
 export const FILE_CLICK_NOTIFICATION_METHOD = 'aws/chat/fileClick'
 export const INLINE_CHAT_REQUEST_METHOD = 'aws/chat/sendInlineChatPrompt'
 export const TAB_BAR_ACTION_REQUEST_METHOD = 'aws/chat/tabBarAction'
+export const SET_METADATA_NOTIFICATION_METHOD = 'aws/chat/setMetadata'
```
volodkevych, line 21: what do you think about naming it 'aws/chat/chatOptionsUpdate'

types/chat.ts
```diff
@@ -281,6 +282,10 @@ export interface ChatUpdateParams {
     data?: TabData
 }
 
+export interface SetMetadataParams extends Partial<ChatMessage> {
```
volodkevych, line 285: Everything is optonal in `ChatMessage`, so do not see much value in this `Partial`.

types/chat.ts
```diff
@@ -281,6 +282,10 @@ export interface ChatUpdateParams {
...
+    content: string
```
volodkevych, line 286: Why do we need `content` if we have `body` inside of `ChatMessage`?

volodkevych: Following `chatOptionsUpdate` idea, what do you think about putting into params:
`tabNotifications?: ChatMessage[]`?

types/chat.ts
volodkevych: We should remember to extend README with protocol whenever we change protocol.

types/chat.ts
volodkevych: Can we please also put explicit explanation why this tab notifications require separare protocol and cannot fit into base chat messages?

types/chat.ts
vidkhath: Its possible that we introduce any required fields in the future right (although that would be backwards incompatible)

types/chat.ts
vidkhath: I think its okay, we could call it chatOptionsUpdate but it might be confused with chatOptions and I am not sure if functionality is the same as chatOptions.

vidkhath: and good catch, I will update README

types/chat.ts
vidkhath: The idea of using `content` was because we did not want to specifically specify `profileName` as a field. I am happy to change it to body of the `ChatMessage`. I would refrain from calling it `tabNotifications` because this notification is tab agnostic. Maybe just `notification` or `chatNotification` works? And I will add an explanation , thanks.

runtimes/README.md
```diff
@@ -188,6 +188,8 @@ The runtime supports chat by default
...
 | Send conversation or conversation action click event. If no action is provided, the default action is "open". | `aws/chat/conversationClick`          | `ConversationClickParams`          | [Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) Client to Server | `ConversationClickResult`             |
+| Send server-initiated chat metadata notification for developer profile changes. | `aws/chat/optionsUpdate`          | `ChatOptionsUpdateParams`          | [Notification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#notificationMessage) Server to Client | n/a             |
```
volodkevych, line 191: The idea is that the params should be extensible with more chat options in the future, notification for developer profile changes is the real use case for now.

types/chat.ts
```diff
@@ -281,6 +282,14 @@ export interface ChatUpdateParams {
     data?: TabData
 }
 
+/**
+ * Server-initiated chat metadata updates.
+ * Processes changes of developer profiles.
+ */
+export interface ChatOptionsUpdateParams {
+    chatNotification?: ChatMessage[]
```
volodkevych, line 290: nit: given it's an array, it should be in plural

runtimes/README.md
volodkevych, line 191: typo: 'aws/chat/chatOptionsUpdate'

types/chat.ts
volodkevych: > I am not sure if functionality is the same as chatOptions
We could add the same `chatNotifications` to chat options as well - so servers would be able to set it either once and/or update based on server logic. From chat-client point of view, it should not be much more work.

types/chat.ts
```diff
@@ -190,6 +191,12 @@ export interface ChatOptions {
      * Server signals to Chat Client support of Chat export feature.
      */
     export?: boolean
+
+    /*
+        Server signals to Chat Client support of Chat notifications.
+        Currently only supports chat notifications for developer profiles.
```
volodkevych, line 197: Why is it? The type is just `ChatMessage` - we could put any text into the `body` there. "developer profiles" - this is how we plan to use it, but I do not think the protocol limits us to this use case.

volodkevych (approved):

volodkevych (approved):

PR #436, opened 2025-04-14
Add initialization configuration options for file indexing

taldekar:
## Problem
Local project context needs to be configurable within different IDEs with rules for files to exclude from indexing, as well as supported languages.

## Solution
Add a `ContextConfiguration` parameter to the `AWSInitializationOptions` interface to be provided during initialization that will determine how local project context will index the workspace. Additional user defined configuration will be available through the `workspace/configuration` call.

### DIFF
--- a/runtimes/protocol/lsp.ts
+++ b/runtimes/protocol/lsp.ts
@@ -48,10 +48,10 @@ export class AutoParameterStructuresProtocolRequestType<P, R, PR, E, RO>
 }
 
 /**
- * Configuration interface for setting up local context file indexing.
- * Controls what files are indexed, size limitations, and indexing behavior.
+ * Configuration options for file indexing in local project context.
+ * Controls what files are indexed.
  */
-export interface ContextConfiguration {
+export interface WorkspaceIndexConfiguration {
     /**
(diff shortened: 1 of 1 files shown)


runtimes/protocol/lsp.ts
```diff
@@ -117,6 +135,10 @@ export interface AWSInitializationOptions {
             showSaveFileDialog?: boolean
         }
     }
+    /**
+     * Local project context file indexing configuration options
+     */
+    contextConfiguration?: ContextConfiguration
```
viktorsaws, line 141: It may be better to name this more explicitly, like `workspaceIndexingConfiguration`

taldekar: @justinmk3 I wasn't aware that there are more than one type of contexts, so in my mind this was more of a global context. Do you think that a "global" context would benefit from the ignoreFiles and fileExtensions fields? If not, I can make it more specific to file indexing.

justinmk3: It's not a well-defined concept, but your latest change which makes `workspaceIndexConfiguration` a property of the "general" context is a way to avoid needing to solve that now :)

breedloj (approved):

PR #441, opened 2025-04-16
fix: revert the version of jose to 5.9.6

pras0131:
## Problem
The language-servers is crashing because jose v6.x is now purely an ESM package and no longer supports CommonJS require(). Since the project is configured to use CommonJS (as defined in the tsconfig.json), this creates a compatibility issue. I saw the below error when i tried to run the server on my local using jose v6.0.10

```
/Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/auth/standalone/encryption.js:8
const jose_1 = require("jose");
               ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/node_modules/jose/dist/webapi/index.js from /Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/auth/standalone/encryption.js not supported.
Instead change the require of index.js in /Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/auth/standalone/encryption.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/auth/standalone/encryption.js:8:16)
    at Object.<anonymous> (/Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/standalone.js:53:22)
    at Object.<anonymous> (/Users/{user}/Workspaces/language-servers/node_modules/@aws/language-server-runtimes/runtimes/index.js:4:20)
    at Object.<anonymous> (/Users/{user}/Workspaces/language-servers/app/aws-lsp-codewhisperer-runtimes/out/token-standalone.js:3:20) {
  code: 'ERR_REQUIRE_ESM'
}

```
Attaching the screenshot below from https://www.npmjs.com/package/jose, showing only ESM being supported for v6+

Screenshot 2025-04-16 at 11 15 28

## Solution
Reverting the version of jose back to 5.9.6

## License

### DIFF
--- a/runtimes/package.json
+++ b/runtimes/package.json
@@ -48,7 +48,7 @@
         "aws-sdk": "^2.1692.0",
         "axios": "^1.8.4",
         "hpagent": "^1.2.0",
-        "jose": "^6.0.10",
+        "jose": "^5.9.6",
         "mac-ca": "^3.1.1",
         "rxjs": "^7.8.2",
         "vscode-languageserver": "^9.0.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1398,155 +1398,6 @@
                 "@opentelemetry/api": "^1.3.0"
             }
         },
-        "node_modules/@opentelemetry/exporter-metrics-otlp-http": {
-            "version": "0.200.0",
-            "resolved": "https://registry.npmjs.org/@opentelemetry/exporter-metrics-otlp-http/-/exporter-metrics-otlp-http-0.200.0.tgz",
-            "integrity": "sha512-5BiR6i8yHc9+qW7F6LqkuUnIzVNA7lt0qRxIKcKT+gq3eGUPHZ3DY29sfxI3tkvnwMgtnHDMNze5DdxW39HsAw==",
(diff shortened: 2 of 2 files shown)


viktorsaws (approved):

ege0zcan (approved):

runtimes/package.json
```diff
@@ -48,7 +48,7 @@
         "aws-sdk": "^2.1692.0",
         "axios": "^1.8.4",
         "hpagent": "^1.2.0",
-        "jose": "^6.0.10",
+        "jose": "^5.9.6",
```
ege0zcan, line 51: we should let dependabot know that we don't want major version updates for jose library

pras0131: right, I'll update the configuration of dependabot.yaml as a follow up to this one.

PR #452, opened 2025-04-20
chore(chat-client-ui-types): add pair programming card acknowledgement command

jpinkney-aws:
## Problem
we need a way to dismiss the pair programming card used in agentic chat

## Solution
when the "X" is pressed on the pair programming card it will send a pairProgrammingAcknowledged event to VSCode

## License


aditya169 (approved):

volodkevych (approved):

chat-client-ui-types/src/uiContracts.ts
```diff
@@ -44,6 +45,7 @@ export type UiMessageCommand =
     | typeof CHAT_OPTIONS
     | typeof COPY_TO_CLIPBOARD
     | typeof DISCLAIMER_ACKNOWLEDGED
+    | typeof PAIR_PROGRAMMING_ACKNOWLEDGED
```
volodkevych, line 48: What will extension do when it gets this event?

jpinkney-aws: store that pair programming has been acknowledged so the next time VSCode launches the chat client it won't be shown

PR #463, opened 2025-04-21
feat: add stopResponse ui event

jpinkney-aws:
## Problem
for agentic chat we need a way for the chat client to tell VSCode that we should stop the chat request stream

## Solution
add it

## Other notes
The implementation of how this event would be used is:
- send a notification from chat-client to VSCode
- cancel the token for the active chat request stream
- replicate the equivalent of [this code](https://github.com/aws/aws-toolkit-vscode/blob/feature/agentic-chat/packages/core/src/codewhispererChat/controllers/chat/controller.ts#L415) in flare

## License

### DIFF
--- a/chat-client-ui-types/src/uiContracts.ts
+++ b/chat-client-ui-types/src/uiContracts.ts
@@ -106,6 +106,15 @@ export interface ChatPromptOptionAcknowledgedMessage {
     params: ChatPromptOptionAcknowledgedParams
 }
 
+export interface StopResponseParams {
+    tabId: string
+}
+
+export interface StopResponseMessage {
+    command: typeof STOP_RESPONSE
+    params: StopResponseParams
+}
+
 export interface ErrorParams {
     tabId: string
     triggerType?: TriggerType
--- a/chat-client-ui-types/src/uiContracts.ts
+++ b/chat-client-ui-types/src/uiContracts.ts
@@ -27,6 +27,7 @@ export const GENERIC_COMMAND = 'genericCommand'
 export const CHAT_OPTIONS = 'chatOptions'
 export const DISCLAIMER_ACKNOWLEDGED = 'disclaimerAcknowledged'
 export const CHAT_PROMPT_OPTION_ACKNOWLEDGED = 'chatPromptOptionAcknowledged'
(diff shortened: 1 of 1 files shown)


types/chat.ts
```diff
@@ -287,6 +288,10 @@ export interface InsertToCursorPositionParams {
     totalCodeBlocks?: number
 }
 
+export interface StopChatResponseParams {
+    tabId: string
```
kmile, line 292: Is this enough to know which request or prompt to cancel?

Hweinstock (approved):
Should we use `stopChatRequest` since its going from client to server? However, if this is consistent with other events then this makes sense.

kmile (approved):
Is there a reason LSP request cancellation isn't possible here?

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#cancelRequest

This implementation looks like it's more specific, cancel any open model invocations for a chat tab independent of whether there's an open `onChatPromp` request or not.

chat-client-ui-types/src/uiContracts.ts
```diff
@@ -27,6 +27,7 @@ export const GENERIC_COMMAND = 'genericCommand'
 export const CHAT_OPTIONS = 'chatOptions'
 export const DISCLAIMER_ACKNOWLEDGED = 'disclaimerAcknowledged'
 export const CHAT_PROMPT_OPTION_ACKNOWLEDGED = 'chatPromptOptionAcknowledged'
+export const STOP_RESPONSE = 'stopResponse'
```
jpinkney-aws, line 30: does it make sense for this to be STOP_RESPONSE to be more generic? or should it be stopChatResponse which is closer to mynah UI's name for the event that will send it?

Hweinstock: Is there a use case for this outside of stopping a chat response? If not, I think mirroring mynah UI could be helpful?

jpinkney-aws: Not sure, I've just updated it to mirror mynah UI. I guess since this is the chat clients package it makes sense that it would stopChatResponse anyways

Hweinstock (approved):

PR #484, opened 2025-04-25
feat: add autoSubmit param to SendToPromptParams

jpinkney-aws:
## Problem
To support `aws.amazonq.ExplainIssue` in security scan for VSCode we need to have a way to send a generic message to the chat client and automatically submit that message

## Solution
Add an autoSubmit, prompt param to SendToPromptParams. This option will be disabled by default. Only when the IDE explicitily indicates that the user intends for this to automatically send will it auto submit

## License

### DIFF
--- a/chat-client-ui-types/src/uiContracts.ts
+++ b/chat-client-ui-types/src/uiContracts.ts
@@ -7,6 +7,7 @@ import {
     OpenTabResult,
     GET_SERIALIZED_CHAT_REQUEST_METHOD,
     GetSerializedChatResult,
+    ChatPrompt,
 } from '@aws/language-server-runtimes-types'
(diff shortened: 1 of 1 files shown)


chat-client-ui-types/src/uiContracts.ts
```diff
@@ -66,6 +66,8 @@ export type UiMessageParams =
 export interface SendToPromptParams {
     selection: string
     triggerType: TriggerType
+    prompt?: string
```
volodkevych, line 69: in't `selection` for the same?

volodkevych: Or it's just for the selected code?

volodkevych (approved):

volodkevych (approved):

PR #487, opened 2025-04-29
feat(types): add isEdit flag for next edit suggestion support

floralph:
Implements the type changes needed for NEP edit suggestions in the language-server-runtimes repository.

### DIFF
--- a/types/inlineCompletionWithReferences.ts
+++ b/types/inlineCompletionWithReferences.ts
@@ -11,11 +11,11 @@ export type InlineCompletionItemWithReferences = InlineCompletionItem & {
 
     /**
      * POC-NEP: Flag to indicate if this is an edit suggestion rather than an inline completion.
-     * When true, the client should render this as an edit suggestion using the VSCode proposed API.
-     * This is used by the Next Edit Prediction feature to distinguish between inline completions
-     * and edit suggestions.
+     * This uses the same name as the proposed VSCode API's isInlineEdit property.
+     * When the VSCode API is finalized, this property can be removed as it will be
+     * part of the standard InlineCompletionItem interface.
      */
-    isEdit?: boolean
+    isInlineEdit?: boolean
 
     references?: {
         referenceName?: string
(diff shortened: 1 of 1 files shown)


types/inlineCompletionWithReferences.ts
```diff
@@ -9,6 +9,14 @@ export type InlineCompletionItemWithReferences = InlineCompletionItem & {
      */
     itemId: string
 
+    /**
+     * POC-NEP: Flag to indicate if this is an edit suggestion rather than an inline completion.
+     * When true, the client should render this as an edit suggestion using the VSCode proposed API.
+     * This is used by the Next Edit Prediction feature to distinguish between inline completions
+     * and edit suggestions.
+     */
+    isEdit?: boolean
```
floralph, line 18: Why aren't we using the isInlineEdit name from the proposed API that will be on InlineCompletionItem in the future.  Once it is, we can remove it from here.  Update the comments here to make that clear.

types/inlineCompletionWithReferences.ts
```diff
@@ -58,6 +66,12 @@ export interface InlineCompletionStates {
      * Recommendation was filtered out on the client-side and marked as discarded.
      */
     discarded: boolean
+
+    /**
+     * POC-NEP: Flag to indicate if this was an edit suggestion rather than an inline completion.
+     * This is used by the Next Edit Prediction feature for telemetry and to improve future suggestions.
+     */
+    isEdit?: boolean
```
floralph, line 74: This isn't a completion state and doesn't belong here.  It looks like InlineCompletionStates is used in LogInlineCompletionSessionResultsParams below.  I think we shouldn't pass the flag for edit back and forth between the client and server here at all.  When the server generates a sessionId, it should keep up with the state of whether the associated suggestion was an edit or not.  That way it's entirely a server implementation detail and reduces complexity.

floralph:
I've updated the code based on your feedback:\n\n1. Changed the property name from  to  to match the VSCode proposed API name\n2. Updated the comment to clarify that this property can be removed once the VSCode API is finalized\n3. Removed the  property from  as it's not a completion state and should be tracked by the server instead\n\nThese changes make the code more aligned with the VSCode API and simplify the client-server communication.

PR #489, opened 2025-04-30
feat: change getConfigurationFromServer to fetch from all servers and aggregate

pras0131:
## Problem
There is a request to return **awsServerCapabilities** from **aws/getConfigurationFromServer**. Due to the way VS is structured, there is no way to get the response payload from standard InitializeResult extension.

## Solution
**getConfigurationFromServer** implementation has been changed to fetch the configurations from all servers instead of first capable server, as different servers might have their individual awsServerCapabilities defined and we want to return all of those. Hence, there is a logic to fetch from all servers and then merge the responses while returning back. 

There is an extra check `
        if (results.length <= 1) {
            return results.pop()
        }` to make sure we don't runt unnecessarily run reduceRight function, when either undefined or the only value in the array needs to be returned. Also, reduceRight gives an error on an empty array argument. Existing mergeObjects has been re-used which has been tested and suffices the current use case with just one extra condition for primitive types. Post this PR merge and runtimes version update in language-servers, respective changes to return server capabilities will be made in individual servers (Making those changes early and merging them could be problematic as runtimes behaviour might cause issues otherwise due to the current nature of fetching configuration from first possible capable server).

## License


viktorsaws:
`awsServerCapabilities` is static object returned from each server at initialization handshake. With this change, are you expecting that every Server will have to implement `aws/getConfigurationFromServer` with `awsServerCapabilities` section in it? This maybe not scale very well, or we would need to enforce this somehow on every server.

Since runtime already processes and aggregates `initializeResults` from each server, it can be easier to cache that object in lspRouter and make it to handle `aws.awsServerCapabilities` section for `aws/getConfigurationFromServer` request, instead of polling every Server.

runtimes/runtimes/lsp/router/lspRouter.test.ts
```diff
@@ -372,7 +372,7 @@ describe('LspRouter', () => {
     })
 
     describe('handleGetConfigurationFromServer', () => {
-        it('should return the result from the first server that handles the request', async () => {
+        it('should return the aggregated result from all servers', async () => {
```
viktorsaws, line 375: Test body hasn't change, which means that either implementation does not do what is suppose to do, or this test never worked. I expect it would start failing

pras0131: Test works fine as expected as coincidentally it's an edge case where property is a string and based on how mergeObjects was implemented, I followed a similar pattern to return obj2 property - in our case, string with server2.

pras0131:
> `awsServerCapabilities` is static object returned from each server at initialization handshake. With this change, are you expecting that every Server will have to implement `aws/getConfigurationFromServer` with `awsServerCapabilities` section in it? This maybe not scale very well, or we would need to enforce this somehow on every server.
> 
> Since runtime already processes and aggregates `initializeResults` from each server, it can be easier to cache that object in lspRouter and make it to handle `aws.awsServerCapabilities` section for `aws/getConfigurationFromServer` request, instead of polling every Server.


>With this change, are you expecting that every Server will have to implement `aws/getConfigurationFromServer` with `awsServerCapabilities` section in it?

Yes, I was thinking that every server can return their own version of server capabilities and then we aggregate just like we do for initialize results. On the scaling part, yes we will then have to make sure to have similar structure of `awsServerCapabilities` returned in every servers' initializeResult and getConfigurationFromServer (for aws.key.serverCapabilities` key)

>Since runtime already processes and aggregates `initializeResults` from each server, it can be easier to cache that object in lspRouter and make it to handle `aws.awsServerCapabilities` section for `aws/getConfigurationFromServer` request, instead of polling every Server.

I agree here that it can be made a bit simpler this way. However, then we will have to do special case handling for `aws.q.serverCapabilities` inside `lspRouter:getConfigurationFromServer` in the runtimes package, whereas for other configurations, it will be generic call to getConfigurationFromServer for the first capable server. Doing this I can then avoid even passing `aws.q.serverCapabilities` in the initializeResult as well.

runtimes/runtimes/lsp/router/lspRouter.test.ts
volodkevych: +1, this test is not testing the aggregation logic, just by accidence `server2` contains aggregated sections. We need proper test to check aggregation is working.

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -125,11 +125,34 @@ ${JSON.stringify({ ...result.capabilities, ...result.awsServerCapabilities })}`
...
+            errors.forEach(error => {
+                this.lspConnection.console.log(
+                    `Error in getting server configuration for section ${params.section}: ${error.message}`
+                )
+            })
+            return new ResponseError(ErrorCodes.InternalError, 'Error in getting server configuration', errors)
+        }
+
+        if (results.length <= 1) {
+            return results.pop()
```
volodkevych, line 146: This will return `undefined` if no elements? I think this is not expected.

pras0131: yes..but right now removing these changes as Viktor suggested something simpler where we don't have to do any of these.

runtimes/runtimes/lsp/router/constants.ts
```diff
@@ -0,0 +1 @@
+export const Q_SERVER_CAPABILITES_CONFIGURATION_SECTION = 'aws.q.serverCapabilities'
```
volodkevych, line 1: It's not Q-specific, right?

pras0131: Right, it should be instead aws.serverCapabilities

volodkevych (approved):

PR #492, opened 2025-05-04
Add displayLocation property and inlineEditSupport capability for tab-tab-tab workflow

floralph:
## Problem
The Next Edit Predictions (NEP) feature requires support for the tab-tab-tab workflow, which allows users to navigate through a series of related edit suggestions. However, the current implementation lacks the necessary properties and capability flags to support this workflow.

## Solution
• Added the displayLocation property to InlineCompletionItemWithReferences with a concise comment explaining its purpose for tab-tab-tab workflow navigation
• Updated the comment for isInlineEdit to be more concise and focused on its functionality
• Added the inlineEditSupport capability to the awsClientCapabilities structure under textDocument.inlineCompletionWithReferences

These changes enable:
1. Clients to declare support for inline edit features during initialization
2. Servers to detect this capability and conditionally include the new properties
3. Servers to decide whether to run the NEP code path based on client support

The implementation follows the LSP specification pattern for capability negotiation while maintaining backward compatibility with existing clients.

## Testing
• Verified the changes compile without errors
• Confirmed the capability structure follows LSP conventions
• Validated the property definitions match the requirements from the NativeUI-POC implementation

## License


dk19y (approved):

types/inlineCompletionWithReferences.ts
```diff
@@ -9,6 +9,22 @@ export type InlineCompletionItemWithReferences = InlineCompletionItem & {
      */
     itemId: string
 
+    /**
+     * Flag to indicate this is an edit suggestion rather than a standard inline completion.
+     */
+    isInlineEdit?: boolean
```
volodkevych, line 15: Will we support 1 inline completion and another prediction within the same client? If no, this flag is not needed as client already knows if it enabled edit predictions or no.

volodkevych (approved):

PR #496, opened 2025-05-06
fix: avoid path import on web.

Hweinstock:
## Problem
path is imported to join paths on web. This was assumed to be safe since `path` is polyfilled with `path-browserify` in the webpack config. However, this polyfill requires consumers to add this dependency, which can be undesirable. 

## Solution
- write our own simplified join for unix paths that can be used in the browser case. 
- remove path import. 

## License

### DIFF
--- a/runtimes/runtimes/util/pathUtil.ts
+++ b/runtimes/runtimes/util/pathUtil.ts
@@ -1,3 +1,8 @@
+/**
+ * Simplified version of path.join that can be safely used on web
+ * @param segments
+ * @returns
+ */
 export function joinUnixPaths(...segments: string[]): string {
     // Filter out empty segments and normalize each segment
     const normalizedSegments = segments.filter(Boolean).map(segment => segment.replace(/^\/+|\/+$/g, ''))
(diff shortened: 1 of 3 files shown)


runtimes/runtimes/util/pathUtil.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Simplified version of path.join that can be safely used on web
+ * @param segments
+ * @returns
+ */
+export function joinUnixPaths(...segments: string[]): string {
```
justinmk3, line 6: "unix" in the name will discourage use for windows. maybe we can just add a `TODO: support "\" slashes` here, to avoid callers thinking they need some windows-specific join
```suggestion
export function joinPaths(...segments: string[]): string {
```

runtimes/runtimes/util/pathUtil.ts
```diff
@@ -0,0 +1,28 @@
...
+    // Filter out empty segments and normalize each segment
+    const normalizedSegments = segments.filter(Boolean).map(segment => segment.replace(/^\/+|\/+$/g, ''))
```
justinmk3, line 8: can also replace `\` slashes with `/` for Windows. then the following logic works on all platforms

runtimes/runtimes/util/pathUtil.ts
Hweinstock: This was actually my intention since I don't see a use case for this utility outside of the web environment where we mimic unix paths. Is there a non-web case where we want to use this over `path.join`?

imykhai: Since we need the fix to unblock the web bundling, let's make the changes suggested by @justinmk3 in the followup

imykhai (approved):

PR #499, opened 2025-05-09
test: operational telemetry without mocking sdk

natdurlik:
## Problem
We need more comprehensive tests for operational telemetry.

## Solution
* Implemented tests for `OperationalTelemetryService` using real OpenTelemetry SDK interactions instead of mocks. 
* Set up a local test server to capture and validate OpenTelemetry HTTP exports.
* Enhanced telemetry validation by implementing strict data schema checks on exported endpoints. Any unplanned additions to the telemetry data will now trigger test failures, improving visibility of any changes.
* Removed old mock-based tests.

Replaced `NodeSDK` with direct provider registration for logs and metrics. This change removes unnecessary trace provider registration that was previously included with `NodeSDK`. The new implementation properly handles telemetry service lifecycle by explicitly deregistering providers when telemetry is disabled, avoiding the registration conflicts that occurred when attempting to re-start telemetry service.

## License


runtimes/runtimes/operational-telemetry/telemetry-schemas/telemetry-schema.json
```diff
@@ -1,18 +1,5 @@
 {
     "$schema": "http://json-schema.org/draft-07/schema#",
-    "title": "OperationalTelemetrySchemas",
-    "type": "object",
-    "oneOf": [
```
ege0zcan, line 5: Just to make sure, with this removed we are not losing any extra validation we had right?

Would it be possible to somehow add a test that tries to send a telemetry event that violates the schema and make sure that it is not getting sent (If such a test already exists then ignore this comment)

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
```diff
@@ -96,57 +93,83 @@ export class OperationalTelemetryService implements OperationalTelemetry {
...
         try {
-            await this.sdk?.shutdown()
+            const promises: Promise<unknown>[] = []
+            if (this.loggerProvider) {
+                promises.push(this.loggerProvider.shutdown())
+            }
+            if (this.meterProvider) {
+                promises.push(this.meterProvider.shutdown())
+            }
+            await Promise.all(promises).then(() => {})
```
ege0zcan, line 160: nit: what is the `.then(() => {})` part used for?

runtimes/runtimes/operational-telemetry/operational-telemetry-service.test.ts
```diff
@@ -1,120 +1,230 @@
...
+            // @ts-ignore
+            await instance.shutdownApi()
+        }
+
+        await promisify(server.close.bind(server))()
     })
 
-    afterEach(() => {
-        sinon.restore()
+    function makeMetricsRequestDeterministic(jsonStr: string): string {
```
ege0zcan, line 65: nit, isn't it possible to modify the json of the metrics and logs before we stringify them? This way it looks like we are creating the json objects, stringifying and then doing string operations. If it is possible to just modify the json it would be simpler

runtimes/runtimes/operational-telemetry/operational-telemetry-service.test.ts
```diff
@@ -1,120 +1,230 @@
...
+            ...getOperationalTelemetryConfig(),
+            telemetryOptOut: true,
         })
 
-        it('should shutdown SDK when switching from opt-in to opt-out', () => {
-            const config = { ...defaultConfig, telemetryOptOut: false }
-            const optel = OperationalTelemetryService.getInstance(config)
+        assert(
+            //@ts-ignore
+            OperationalTelemetryService.instance.loggerProvider === null,
```
ege0zcan, line 184: does it make sense to verify the loggerProvider and meterProvider are instantiated when the user is opted-in or would it be redundant given we already test that http requests are leaving the telemetry service?

ege0zcan (approved):

runtimes/runtimes/operational-telemetry/telemetry-schemas/telemetry-schema.json
natdurlik: We currently don't use "OperationalTelemetrySchemas" for anything. We previously had this part because the `generate-types.ts` script wasn't generating all types, but this issue was fixed in a previous PR by adding `--unreachableDefinitions` flag.

If I understood you correctly, we have such tests, but they're not in this repository.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.ts
natdurlik: yes, this is redundant

runtimes/runtimes/operational-telemetry/operational-telemetry-service.test.ts
natdurlik: I agree, we're converting json to a string and then back to json again, but using regular expressions was the easiest way for me to normalize it. The values we want to modify are deep in the json request object so it isn't straightforward to parse it.

runtimes/runtimes/operational-telemetry/operational-telemetry-service.test.ts
```diff
@@ -1,120 +1,230 @@
...
+    }
+
+    it('should send metrics to the configured endpoint', async function () {
+        this.timeout(10000)
+
+        const telemetryService = OperationalTelemetryService.getInstance(getOperationalTelemetryConfig())
+        telemetryService.registerGaugeProvider('heapUsed', () => 12345, 'byte')
+
+        await waitForRequests(1, 5000)
+        assert(receivedRequests.length > 0, 'Should have received at least one request')
```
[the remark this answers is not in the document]
natdurlik: right, current assert is redundant, let me change it to be more strict

runtimes/runtimes/operational-telemetry/operational-telemetry-service.test.ts
natdurlik: I think it's enough to see that we are receiving the requests

francescoopiccoli (approved):

ege0zcan (approved):

ege0zcan (approved):

PR #502, opened 2025-05-09
feat: adding new params for mcp servers buttons

ashishrp-aws:
## Problem

- MCP server UI needs new UI buttons and events

## Solution
- Adding required events for MCP UI events.

## License

### DIFF
--- a/runtimes/protocol/chat.ts
+++ b/runtimes/protocol/chat.ts
@@ -67,6 +67,12 @@ import {
     ButtonClickParams,
     ButtonClickResult,
     BUTTON_CLICK_REQUEST_METHOD,
+    ListMcpServersParams,
+    ListMcpServersResult,
+    LIST_MCP_SERVERS_REQUEST_METHOD,
+    McpServerClickParams,
+    McpServerClickResult,
+    MCP_SERVERS_CLICK_REQUEST_METHOD,
 } from './lsp'
 
 export const chatRequestType = new AutoParameterStructuresProtocolRequestType<
@@ -163,6 +169,23 @@ export const conversationClickRequestType = new AutoParameterStructuresProtocolR
     void,
     void
 >(CONVERSATION_CLICK_REQUEST_METHOD)
+
+// mcp servers
+export const listMcpServersRequestType = new AutoParameterStructuresProtocolRequestType<
+    ListMcpServersParams,
+    ListMcpServersResult,
+    never,
+    void,
+    void
+>(LIST_MCP_SERVERS_REQUEST_METHOD)
+
+export const mcpServerClickRequestType = new AutoParameterStructuresProtocolRequestType<
+    McpServerClickParams,
+    McpServerClickResult,
+    never,
+    void,
+    void
+>(MCP_SERVERS_CLICK_REQUEST_METHOD)
 
 /**
  * The tab bar action request is sent from client to server to execute action from Chat tab bar UI.
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -29,6 +29,9 @@ export const INLINE_CHAT_RESULT_NOTIFICATION_METHOD = 'aws/chat/inlineChatResult
 // history
 export const LIST_CONVERSATIONS_REQUEST_METHOD = 'aws/chat/listConversations'
 export const CONVERSATION_CLICK_REQUEST_METHOD = 'aws/chat/conversationClick'
+//mcpServers
+export const LIST_MCP_SERVERS_REQUEST_METHOD = 'aws/chat/listMcpServers'
+export const MCP_SERVERS_CLICK_REQUEST_METHOD = 'aws/chat/mcpServersClick'
 // export
 export const GET_SERIALIZED_CHAT_REQUEST_METHOD = 'aws/chat/getSerializedChat'
 
@@ -234,6 +237,8 @@ export interface ChatOptions {
      * Chat QuickActions, supported by Server. Chat Client renders and sets up actions handler for registered QuickAction in UI.
      */
     quickActions?: QuickActions
+
+    mcpServers?: boolean
 
     /**
      * Server signals to Chat Client support of conversation history.
@@ -445,6 +450,10 @@ export interface ListConversationsParams {
     filter?: Record<string, FilterValue>
 }
 
+export interface ListMcpServersParams {
+    filter?: Record<string, FilterValue>
+}
+
 export interface ConversationsList {
     header?: { title: string }
     filterOptions?: FilterOption[]
@@ -452,6 +461,8 @@ export interface ConversationsList {
 }
 
 export interface ListConversationsResult extends ConversationsList {}
+
+export interface ListMcpServersResult extends ConversationsList {}
 
 export type ConversationAction = 'delete' | 'export'
 
@@ -461,6 +472,15 @@ export interface ConversationClickParams {
 }
 
 export interface ConversationClickResult extends ConversationClickParams {
+    success: boolean
+}
+
+export interface McpServerClickParams {
+    id: string
+    action?: ConversationAction
+}
+
+export interface McpServerClickResult extends McpServerClickParams {
     success: boolean
 }
 


runtimes/protocol/chat.ts
```diff
@@ -67,6 +67,12 @@ import {
     ButtonClickParams,
     ButtonClickResult,
     BUTTON_CLICK_REQUEST_METHOD,
+    ListMcpServersParams,
+    ListMcpServersResult,
+    LIST_MCP_SERVERS_REQUEST_METHOD,
+    McpServerClickParams,
```
bywang56, line 73: So LIST_MCP_SERVERS_REQUEST_METHOD will be used by only when user click the new MCP configuration button to open the Server list?
and for MCP_SERVERS_CLICK_REQUEST_METHOD, when this will be used? is it used for all button clicks in the MCP view? why can't we use the BUTTON_CLICK_REQUEST_METHOD?

ashishrp-aws: Discussed offline...for List MCP we can use alternate button click, which also needs similar expansion to add MCP server details. Since the button is in tab button list. Thought this would be much cleaner and i've already tested it with using listHistoryButton

types/chat.ts
```diff
@@ -464,6 +475,15 @@ export interface ConversationClickResult extends ConversationClickParams {
     success: boolean
 }
 
+export interface McpServerClickParams {
+    id: string
+    action?: ConversationAction
```
avi-alpert, line 480: wouldn't you need your own Mcp Actions?

ashishrp-aws: Yes...but wasn't able to test it yet. Will update this. I'm having issues with my local link from lsr to ls. so setup a PR for only the stuff i tested.

runtimes/protocol/chat.ts
avi-alpert: https://github.com/aws/language-server-runtimes/commit/a4fcef6459f4224a6bf16d3ce6e88ef326fec3aa

See above PR for adding the history related methods (ListConversations & ConversationClick)

I think it makes sense to have dedicated List & Click methods for MCP. The existing button_click i think is for buttons in the chat itself

ashishrp-aws: Agreed. In that case, we can go ahead with this.

types/chat.ts
```diff
@@ -453,6 +462,8 @@ export interface ConversationsList {
 
 export interface ListConversationsResult extends ConversationsList {}
 
+export interface ListMcpServersResult extends ConversationsList {}
```
volodkevych, line 465: nit: it's a bit weird that `ListMcpServersResult extends ConversationsList`. We could rename `ConversationsList` if it's reusable across different methods.

types/chat.ts
volodkevych: How will 'export' action be used for MCP servers?

types/chat.ts
ashishrp-aws: I'm going to rewrite this...because i wasn't able to fully test the end to end for list MCP severs. It will need more info then conversationsList

ashishrp-aws: Will do a fast followup

types/chat.ts
```diff
@@ -29,6 +29,9 @@ export const INLINE_CHAT_RESULT_NOTIFICATION_METHOD = 'aws/chat/inlineChatResult
 // history
 export const LIST_CONVERSATIONS_REQUEST_METHOD = 'aws/chat/listConversations'
 export const CONVERSATION_CLICK_REQUEST_METHOD = 'aws/chat/conversationClick'
+//mcpServers
+export const LIST_MCP_SERVERS_REQUEST_METHOD = 'aws/chat/listMcpServers'
+export const MCP_SERVERS_CLICK_REQUEST_METHOD = 'aws/chat/mcpServersClick'
```
tsmithsz, line 34: Shouldn't it be `mcpServerClick`?

laileni-aws: Can be a fast followUp!

laileni-aws (approved):

types/chat.ts
```diff
@@ -453,6 +462,57 @@ export interface ConversationsList {
...
+        description: string
+    }
+}
+
+export interface DetailedListItemStatus {
+    icon?: string
+    text?: string
+}
+
+export interface DetailedListItem {
```
avi-alpert, line 487: Instead of copying the detailed list item types from mynah-ui, you should create new types that are relevant to the method(s) you're adding

types/chat.ts
```diff
@@ -453,6 +462,57 @@ export interface ConversationsList {
...
+export interface DetailedListGroup {
+    groupName?: string
+    children?: DetailedListItem[]
+}
+
+export interface ListMcpServersResult {
+    selectable?: boolean | 'clickable'
+    header?: {
+        title: string
+        status?: DetailedListItemStatus
```
avi-alpert, line 506: Do you need the server to send all of this? Can't some of these be configured in mynahUi.ts?

types/chat.ts
```diff
@@ -453,6 +462,57 @@ export interface ConversationsList {
...
+        description?: string
+        actions?: DetailedListAction[]
+    }
+    textDirection?: 'row' | 'column'
```
volodkevych, line 510: Why would server need to set this?

types/chat.ts
```diff
@@ -453,6 +462,57 @@ export interface ConversationsList {
...
+    actions?: DetailedListAction[]
+}
+
+export interface DetailedListGroup {
+    groupName?: string
+    children?: DetailedListItem[]
+}
+
+export interface ListMcpServersResult {
+    selectable?: boolean | 'clickable'
```
volodkevych, line 503: Why is this needed in protocol?

types/chat.ts
volodkevych: I suggest to start with bare minimum and extend this when you see property real needs to be set by server.

runtimes/README.md
```diff
@@ -189,6 +189,8 @@ The runtime supports chat by default
...
 | Send conversation or conversation action click event. If no action is provided, the default action is "open". | `aws/chat/conversationClick`          | `ConversationClickParams`          | [Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) Client to Server | `ConversationClickResult`             |
+| Send request to list the MCP servers available: all or based on filter if provided. Similar to conversations, the filter in the request is a map of filter option ids to corresponding values. | `aws/chat/listMcpServers`          | `ListMcpServersParams`          | [Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#requestMessage) Client to Server | `ListMcpServersResult`             |
```
ege0zcan, line 192: Is MCP interactions only going to be limited to chat? I wonder if we should use `aws/chat` for MCP related stuff or maybe consider creating a separte command path like `aws/mcp`

ege0zcan: Based on this PR it looks intentional and mcp is very much tied to chat UI and chat protocol. 

The list of available mcp servers during initialization will be passed as chatOptions and selecting mcp servers is expected to happen through the chat UI. I just want to raised this coupling so that we are aware of this and make sure this aligns with the long term vision

ege0zcan:
As Avi and Oksana suggested, I suggest to start with a minimal working approach for the protocol and only adding new fields to params when we see a definite need for new fields. Creating the protocol as a 1:1 mapping from the UI means we leak UI details to the protocol and having extra fields in the protocol are hard to remove once they are published

runtimes/README.md
laileni-aws: As per todays requirement, this is limited to chat.

ashishrp-aws: yes limited to chat as of right now.

types/chat.ts
ashishrp-aws: Removed this

types/chat.ts
ashishrp-aws: Removed this.

types/chat.ts
ashishrp-aws: Removed this object

PR #507, opened 2025-05-13
feat: emit telemetry event to lsp client on standalone process crash

viktorsaws:
## Problem
No telemetry is reported on standalone process crash.

## Solution
Emit `runtime_processCrash` telemetry event to LSP client when unhandledException is detected with `uncaughtExceptionMonitor` event. Process crash event will include trimmed stack trace and exception origin information.

Proposed event format:

```
interface ErrorTelemetryEvent {
    name: 'runtime_processCrash'
    result: 'Failed'
    data: {
        // As reported to uncaughtExceptionMonitor, can either be 'uncaughtException' or 'unhandledRejection'.
        origin: string
        // Top 10 frames of stack trace, as stack trace from compiled language server contains full minified source code.
        stack: string
    }
    errorData: {
        reason: string
    }
}
```

Example of telemetry event

```
[Trace - 5:46:43 PM] Received notification 'telemetry/event'.
Params: {
    "name": "runtime_processCrash",
    "result": "Failed",
    "data": {
        "origin": "uncaughtException",
        "stack": "Error: Runtime: Process crash timeout reached\n    at Timeout._onTimeout (/Users/vshcherb/Projects/Flare/language-servers/node_modules/@aws/language-server-runtimes/runtimes/standalone.js:152:15)\n    at listOnTimeout (node:internal/timers:581:17)\n    at process.processTimers (node:internal/timers:519:7)"
    },
    "errorData": {
        "reason": "Error: Runtime: Process crash timeout reached"
    }
}
```

## License

### DIFF
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -55,6 +55,7 @@ import {
     SDKClientConstructorV2,
     SDKClientConstructorV3,
     SDKInitializator,
+    MetricEvent,
 } from '../server-interface'
 import { Auth } from './auth'
 import { EncryptedChat } from './chat/encryptedChat'
@@ -89,9 +90,40 @@ if (checkAWSConfigFile()) {
     process.env.AWS_SDK_LOAD_CONFIG = '1'
 }
 
-process.on('uncaughtExceptionMonitor', err => {
-    console.error('Uncaught Exception:', err.message)
-})
+// Set up crash monitoring to emit telemetry before process exits
+function setupCrashMonitoring(telemetryEmitter?: (metric: MetricEvent) => void) {
+    function getTopStackFrames(err: Error): string | undefined {
+        if (!err.stack) {
+            return undefined
+        }
+
+        const stackLines = err.stack.split('\n')
+        const topStackFrames = stackLines.slice(0, 10).join('\n')
+        return topStackFrames
+    }
+
+    process.on('uncaughtExceptionMonitor', (err, origin) => {
+        console.error('Uncaught Exception:', err.message)
+
+        if (telemetryEmitter) {
+            try {
+                telemetryEmitter({
+                    name: 'runtime_processCrash',
+                    result: 'Failed',
+                    data: {
+                        origin: origin,
+                        stack: getTopStackFrames(err),
+                    },
+                    errorData: {
+                        reason: err.toString(),
+                    },
+                })
+            } catch (telemetryError) {
+                console.error('Failed to emit telemetry for uncaught exception:', telemetryError)
+            }
+        }
+    })
+}
 
 /**
  * The runtime for standalone LSP-based servers.
@@ -176,6 +208,9 @@ export const standalone = (props: RuntimeProps) => {
             emitMetric: metric => lspConnection.telemetry.logEvent(metric),
             onClientTelemetry: handler => lspConnection.onNotification(telemetryNotificationType.method, handler),
         }
+
+        // Set up crash monitoring with telemetry
+        setupCrashMonitoring(telemetry.emitMetric)
 
         // Set up the workspace sync to use the LSP Text Document Sync capability
         const workspace: Workspace = {


runtimes/runtimes/standalone.ts
```diff
@@ -89,9 +90,40 @@ if (checkAWSConfigFile()) {
...
+        const topStackFrames = stackLines.slice(0, 10).join('\n')
+        return topStackFrames
+    }
+
+    process.on('uncaughtExceptionMonitor', (err, origin) => {
+        console.error('Uncaught Exception:', err.message)
+
+        if (telemetryEmitter) {
+            try {
+                telemetryEmitter({
```
Hweinstock, line 110: For this metric to be emitted from toolkits, I believe it needs to be added [here](https://github.com/aws/aws-toolkit-common/blob/main/telemetry/definitions/commonDefinitions.json). The fields should also be defined as types there, but I think there are ones you can reuse. 

This will also require bumping the version of the telemetry package in vscode [here](https://github.com/aws/aws-toolkit-vscode/blob/8e85476708e0d2e9467cc7afe13a42661de78f1d/package.json#L44) once that change is in (as well as other IDES).

runtimes/runtimes/standalone.ts
```diff
@@ -89,9 +90,40 @@ if (checkAWSConfigFile()) {
...
+                    name: 'runtime_processCrash',
+                    result: 'Failed',
+                    data: {
+                        origin: origin,
+                        stack: getTopStackFrames(err),
```
Hweinstock, line 115: AFAIK we usually don't include stack since its very noisy. We try to extract relevant information from the error itself and add it as metadata fields.

awschristou:
The shape should use the common structure, and global fields defined in https://github.com/aws/aws-toolkit-common/blob/main/telemetry/telemetryformat.md

I think this becomes:
```
interface ErrorTelemetryEvent {
    name: 'runtime_processCrash'
    result: 'Failed'
    reason: string 
    reasonDesc: string // put stack trace here
    errorCode: string // put origin here (this isn't in the .md but pretty sure this is global)
}

```

runtimes/runtimes/standalone.ts
viktorsaws: Good call, I'd remove stack trace from telemetry and print it to log instead.

awschristou: I disagree with @Hweinstock comments. In the VS extension, we include stack traces as one of the error data fields (the toolkit has one that isn't modelled in the language server's ErrorData type). This has been helpful for diagnosing certain classes of errors. To include it or not is a judgement call the Flare codebase maintainers need to make.

runtimes/runtimes/standalone.ts
awschristou: Can you update the PR Description with a sample of the latest changes?

awschristou (approved):

runtimes/runtimes/standalone.ts
```diff
@@ -89,9 +90,37 @@ if (checkAWSConfigFile()) {
     process.env.AWS_SDK_LOAD_CONFIG = '1'
 }
 
-process.on('uncaughtExceptionMonitor', err => {
-    console.error('Uncaught Exception:', err.message)
-})
+// Set up crash monitoring to emit telemetry before process exits
+function setupCrashMonitoring(telemetryEmitter?: (metric: MetricEvent) => void) {
```
volodkevych, line 94: Why would `telemetryEmitter` be undefined?

volodkevych (approved):

runtimes/runtimes/standalone.ts
```diff
@@ -89,9 +90,37 @@ if (checkAWSConfigFile()) {
...
+    process.on('uncaughtExceptionMonitor', (err, origin) => {
+        console.error('Uncaught Exception:', err.message, getTopStackFrames(err))
+
+        if (telemetryEmitter) {
+            try {
+                telemetryEmitter({
+                    name: 'runtime_processCrash',
+                    result: 'Failed',
+                    errorData: {
+                        reason: err.toString(),
```
justinmk3, line 114: `reason` for a "code" or name (`(err as any).code ?? (err as any).name`).

`reasonDesc` is for details. And should use `getTelemetryReasonDesc` to avoid PII https://github.com/aws/language-servers/blob/a592b8b0dce29b485ab79def373e37ecec04b5ac/server/aws-lsp-codewhisperer/src/shared/utils.ts#L174

PR #509, opened 2025-05-14
feat: support using rootUri for workspaceFolders and handle workspaceChangeEvent

shruti0085:
## Description
Currently in a lot of downstream servers, the `workspace.getWorkspaceFolder` method is used for path resolution. This only relied on the `initializeParams.workspaceFolder` value which did not take into account workspace folder changes(did not listen to `didChangeWorkspaceFolders`) that could have happened during the session.
In addition with VS client as mentioned in https://github.com/aws/language-servers/pull/1210, the workspaceFolder value is not set(more details in that PR) however it does specify that information via `rootUri` and `rootPath`. The utility class from that PR has now been moved centrally to this repo.
A new `getAllWorkspaceFolders()` call has been added as a common utility that servers downstream can use as single source of truth to get latest information of workspaceFolders available to the lsp server.
A follow-up PR to language-servers repo will utilize this change.

## License

### DIFF
--- a/runtimes/runtimes/base-runtime.ts
+++ b/runtimes/runtimes/base-runtime.ts
@@ -124,8 +124,8 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
     const workspace: Workspace = {
         getTextDocument: async uri => documents.get(uri),
         getAllTextDocuments: async () => documents.all(),
-        getWorkspaceFolder: _uri =>
-            lspRouter.clientInitializeParams!.workspaceFolders && lspRouter.clientInitializeParams!.workspaceFolders[0],
+        getWorkspaceFolder: _uri => lspRouter.getAllWorkspaceFolders() && lspRouter.getAllWorkspaceFolders()[0],
+        getAllWorkspaceFolders: () => lspRouter.getAllWorkspaceFolders(),
         fs: {
             copyFile: (_src, _dest, _options?) => Promise.resolve(),
             exists: _path => Promise.resolve(false),
--- a/runtimes/runtimes/lsp/router/initializeUtils.ts
+++ b/runtimes/runtimes/lsp/router/initializeUtils.ts
@@ -1,0 +1,32 @@
+import { InitializeParams, WorkspaceFolder } from 'vscode-languageserver-protocol'
+import path from 'path'
+import { URI } from 'vscode-uri'
+import { RemoteConsole } from 'vscode-languageserver'
+
+export function getWorkspaceFoldersFromInit(params?: InitializeParams, console?: RemoteConsole): WorkspaceFolder[] {
+    if (!params) {
+        return []
+    }
+
+    if (params.workspaceFolders && params.workspaceFolders.length > 0) {
+        return params.workspaceFolders
+    }
+    try {
+        const getFolderName = (parsedUri: URI) => path.basename(parsedUri.fsPath) || parsedUri.toString()
+
+        if (params.rootUri) {
+            const parsedUri = URI.parse(params.rootUri)
+            const folderName = getFolderName(parsedUri)
+            return [{ name: folderName, uri: params.rootUri }]
+        }
+        if (params.rootPath) {
+            const parsedUri = URI.parse(params.rootPath)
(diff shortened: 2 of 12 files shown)


runtimes/server-interface/workspace.ts
```diff
@@ -19,6 +19,7 @@ export type Workspace = {
     getTextDocument: (uri: string) => Promise<TextDocument | undefined>
     getAllTextDocuments: () => Promise<TextDocument[]>
     getWorkspaceFolder: (uri: string) => WorkspaceFolder | null | undefined
+    getAllWorkspaceFolders: () => WorkspaceFolder[]
```
justinmk3, line 22: just make it plural. helps discoverability because it will appear next to the singular case.
```suggestion
    getWorkspaceFolders: () => WorkspaceFolder[]
```

justinmk3, line 22: just make it plural. helps discoverability because it will appear next to the singular case. docstrings could be used for both of these cases, meanwhile...


```suggestion
    /** Gets the current(?) workspace folder as decided by ... */
    getWorkspaceFolder: (uri: string) => WorkspaceFolder | null | undefined
    /** Gets all workspace folders known to <something> */
    getWorkspaceFolders: () => WorkspaceFolder[]
```

shruti0085: Yes I followed the existing pattern. I will add the doc strings for clarity

runtimes/runtimes/base-runtime.ts
```diff
@@ -124,8 +124,8 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
     const workspace: Workspace = {
         getTextDocument: async uri => documents.get(uri),
         getAllTextDocuments: async () => documents.all(),
-        getWorkspaceFolder: _uri =>
-            lspRouter.clientInitializeParams!.workspaceFolders && lspRouter.clientInitializeParams!.workspaceFolders[0],
+        getWorkspaceFolder: _uri => lspRouter.getAllWorkspaceFolders() && lspRouter.getAllWorkspaceFolders()[0],
+        getAllWorkspaceFolders: () => lspRouter.getAllWorkspaceFolders(),
```
ege0zcan, line 128: I think the addition of `getAllWorkspaceFolders` makes this PR more of a feature rather than a fix

I am also noticing that none of the `Workspace` functionality is documented in the  readme, maybe we can start to fix things by documenting `getWorkspaceFolder` and `getAllWorkspaceFolders` as part of this PR or as a followup

runtimes/runtimes/lsp/router/initializeUtils.ts
```diff
@@ -0,0 +1,32 @@
...
+            return [{ name: folderName, uri: params.rootUri }]
+        }
+        if (params.rootPath) {
+            const parsedUri = URI.parse(params.rootPath)
+            const folderName = getFolderName(parsedUri)
+            return [{ name: folderName, uri: parsedUri.toString() }]
+        }
+        return []
+    } catch (error) {
+        console.log(`Error occurred when determining workspace folders: ${error}`)
```
ege0zcan, line 29: potentially we could use `console.warn` or `console.error` here

ege0zcan (approved):

runtimes/runtimes/lsp/router/initializeUtils.ts
```diff
@@ -0,0 +1,32 @@
+import { InitializeParams, WorkspaceFolder } from 'vscode-languageserver-protocol'
+import path from 'path'
+import { URI } from 'vscode-uri'
+import { RemoteConsole } from 'vscode-languageserver'
+
+export function getWorkspaceFoldersFromInit(console: RemoteConsole, params?: InitializeParams): WorkspaceFolder[] {
```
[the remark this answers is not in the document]
shruti0085: That's fair. I added it because of the edge cases as I was writing the tests. Currently from the location where this is invoked it should never be null but considering this is a utility call shared, I am inclined to leave it as it.

awschristou (approved):

PR #514, opened 2025-05-17
fix: adding mcp interfaces

laileni-aws:
## Problem
- Missing some fields in `McpServerClickResult` interface for MCP
## Solution
- Added required fields.

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -420,24 +420,23 @@ export interface InlineChatResultParams {
 }
 
 // history
-export type TextBasedFilterOption = {
-    type: 'textarea' | 'textinput' | 'numericinput'
+export interface FilterOptionBase {
     placeholder?: string
     title?: string
     description?: string
     icon?: IconType
 }
 
-export type OptionBasedFilterOption = {
+export type TextBasedFilterOption = FilterOptionBase & {
+    type: 'textarea' | 'textinput' | 'numericinput'
+}
+
+export type OptionBasedFilterOption = FilterOptionBase & {
     type: 'select' | 'radiogroup'
-    placeholder?: string
-    title?: string
-    description?: string
     options: Array<{
         value: string
         label: string
     }>
-    icon?: IconType
 }
 
 export type BaseFilterOption = TextBasedFilterOption | OptionBasedFilterOption
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -421,18 +421,28 @@ export interface InlineChatResultParams {
 
 // history
 export type TextBasedFilterOption = {
-    type: 'textarea' | 'textinput' | 'select' | 'numericinput' | 'radiogroup'
+    type: 'textarea' | 'textinput' | 'numericinput'
     placeholder?: string
     title?: string
     description?: string
-    options?: Array<{
+    icon?: IconType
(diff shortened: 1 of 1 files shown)


types/chat.ts
```diff
@@ -421,8 +421,14 @@ export interface InlineChatResultParams {
...
 export type TextBasedFilterOption = {
-    type: 'textarea' | 'textinput'
+    type: 'textarea' | 'textinput' | 'select' | 'numericinput' | 'radiogroup'
     placeholder?: string
+    title?: string
+    description?: string
+    options?: Array<{
+        value: string
+        label: string
+    }>
```
ashishrp-aws, line 431: `>` symbol?

types/chat.ts
```diff
@@ -500,7 +509,21 @@ export interface McpServerClickParams {
 }
```
ashishrp-aws, line 509: we would need this additon to `McpServerClickParams` here:
```
    title?: string
    optionsValues?: Record<string, string>
```

types/chat.ts
```diff
@@ -421,8 +421,14 @@ export interface InlineChatResultParams {
 
 // history
 export type TextBasedFilterOption = {
```
volodkevych, line 423: 'select' and 'radiogroup' are not text-based filters anymore. Can filter that depend on `options` be a separate type?

laileni-aws: Yeah created a new type
```
export type OptionBasedFilterOption = {
    type: 'select' | 'radiogroup'
    placeholder?: string
    title?: string
    description?: string
    options: Array<{
        value: string
        label: string
    }>
    icon?: IconType
}

export type BaseFilterOption = TextBasedFilterOption | OptionBasedFilterOption

File: types/chat.ts (hunk 2 above)
Comment (laileni-aws) [reply types/chat.ts:509]: Not required for my useCase of `add config for MCP`, Do you want me to add for any other case?

File: types/chat.ts [hunk 4]
```diff
@@ -500,7 +509,21 @@ export interface McpServerClickParams {
 }
 
 export interface McpServerClickResult extends McpServerClickParams {
-    success: boolean
+    filterOptions?: FilterOption[] | null
```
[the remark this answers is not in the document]
laileni-aws: Yes, we are declaring these options on agenticChatController.ts and filtering/selecting the option value in mynahUI.

types/chat.ts
volodkevych: nit: you could define base type from properties like `title` and `icon`

types/chat.ts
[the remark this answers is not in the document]
volodkevych: Do you pass selected filter values through `optionsValues`? It's work to add comments explaining how the filtering flow is expected to work.

volodkevych (approved):

types/chat.ts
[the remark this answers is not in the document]
ashishrp-aws: Yes we pass selected values for filter options. `McpServerClickResult` is providing filter options to MCP list/sheet in agenticChatController.ts/server. we receive them on Mynah. when selected we pass those values to another event under `McpServerClickParams`

PR #521, opened 2025-05-21
fix: adding summary to ChatMessage

laileni-aws:
## Notes
- Adding missing `summary` field to `ChatMessage` for MCP chat summary UX.

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -420,24 +420,23 @@ export interface InlineChatResultParams {
 }
 
 // history
-export type TextBasedFilterOption = {
-    type: 'textarea' | 'textinput' | 'numericinput'
+export interface FilterOptionBase {
     placeholder?: string
     title?: string
     description?: string
     icon?: IconType
 }
 
-export type OptionBasedFilterOption = {
+export type TextBasedFilterOption = FilterOptionBase & {
+    type: 'textarea' | 'textinput' | 'numericinput'
+}
+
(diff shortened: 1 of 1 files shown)


ashishrp-aws (approved):

types/chat.ts
```diff
@@ -149,6 +149,10 @@ export interface ChatMessage {
         title?: string
         content: SourceLink[]
     }
+    summary?: {
+        content?: ChatMessage
+        collapsedContent?: ChatMessage[]
+    } | null
```
ege0zcan, line 155: nit, do we need the `null` option here given summary is already optional?

types/chat.ts
```diff
@@ -149,6 +149,10 @@ export interface ChatMessage {
         title?: string
         content: SourceLink[]
     }
+    summary?: {
+        content?: ChatMessage
+        collapsedContent?: ChatMessage[]
+    }
```
kmile, line 155: Is this intended to be a recursive property or should we `Omit<>` some things?

You could add it to `ChatResult` instead if this belongs more to the `onChatPrompt` request than the actual message data itself. That's a semantic choice and stops recursion as well, so totally up to you.

kmile (approved):

types/chat.ts
laileni-aws: >Is this intended to be a recursive property

I guess it is acc to mynah model: https://github.com/aws/mynah-ui/blob/main/src/static.ts#L403-L407

PR #522, opened 2025-05-21
fix: remove the supports workspace check for didChangeWorkspaceFolder event

shruti0085:
## Problem
PR #509 added a check to validate whether the client supports workspaceFolders before handling the didChangeWorkspaceFolders event. This fails for VS as VS does not send the workspaceFolder value
## Solution
Check has been removed as it is now assumed that whenever this is invoked clients intentionally want to send this event.
This also unblocks VS which sends the workspaceFolder information with rootUri of the initializeParams

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/lspRouter.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.ts
@@ -59,6 +59,7 @@ export class LspRouter {
         lspConnection.workspace.onDidCreateFiles(this.didCreateFiles)
         lspConnection.workspace.onDidDeleteFiles(this.didDeleteFiles)
         lspConnection.workspace.onDidRenameFiles(this.didRenameFiles)
+        // Note: Using raw notification instead of workspace.OnDidWorkspaceChange to avoid downstream VSCode language server errors when workspaceFolders is undefined(clients like VS)
         lspConnection.onNotification(DidChangeWorkspaceFoldersNotification.type, params => {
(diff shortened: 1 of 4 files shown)


awschristou (approved):

runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -58,6 +59,9 @@ export class LspRouter {
         lspConnection.workspace.onDidCreateFiles(this.didCreateFiles)
         lspConnection.workspace.onDidDeleteFiles(this.didDeleteFiles)
         lspConnection.workspace.onDidRenameFiles(this.didRenameFiles)
+        lspConnection.onNotification(DidChangeWorkspaceFoldersNotification.type, params => {
```
shruti0085, line 62: This setup was intentionally used as using the base lspConnection.workspace.OnDidWorkspaceChange invoked the corresponding event in the downstream vscode language server logic that would throw an error if workspaceFolders was undefined. Related code[ here](https://github.com/microsoft/vscode-languageserver-node/blob/d305c8dab76acdfa05b6b6c4ba6c74ac2c978fec/server/src/common/workspaceFolder.ts#L48)
To circumvent that issue, we use the raw notification instead

ege0zcan: Would be good to document this in the code as well

ege0zcan (approved):

PR #530, opened 2025-05-23
feat: add modelId to chatOptionsUpdateParams

ctlai95:
## Problem

Server needs to be able to tell chat client what the latest selected modelId was

## Solution

Add `modelId` to the existing `ChatOptionsUpdateParams`

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -368,6 +368,7 @@ export interface ChatUpdateParams {
  */
 export interface ChatOptionsUpdateParams {
     chatNotifications?: ChatMessage[]
+    modelId?: string
 }
 
 export type FileAction = 'accept-change' | 'reject-change'


types/chat.ts
```diff
@@ -368,6 +368,7 @@ export interface ChatUpdateParams {
  */
 export interface ChatOptionsUpdateParams {
     chatNotifications?: ChatMessage[]
+    modelId?: string
```
shruti0085, line 371: Can you please update the documentation too to explain what this is expected to convey. Currently the [chatNotifications](https://github.com/ctlai95/language-server-runtimes/blob/8e2e0c325e553b8fe2a8e58ba884e370fbcf3f77/types/chat.ts#L258) are specified with logic specific to developer profiles.

PR #553, opened 2025-06-09
feat(runtimes): add auto-discovery for OS proxy settings

tsmithsz:
## Problem
Standalone runtime is only able to use proxies when the user explicitly set the HTTP(S)_PROXY environment variables. 

Users on corporate networks use Transparent proxies configured on Network level. Retrieving the OS certificates trust store is not enough. As a result of this, customers requests to Q Agentic Chat fail as they do not pass through their corporate proxy.

## Solution

We embedded automatic system‑proxy discovery into the standalone language‑server by adding the lightweight [os‑proxy-config](https://www.npmjs.com/package/os-proxy-config?activeTab=readme) package and a tiny synchronous shim that runs it in a child Node process when no HTTP(S)_PROXY env‑vars are set. The shim was used to keep the method synchronous, so that the async changes would not ripple through every SDK-client factory and the many call-sites where these methods are used.

The server now transparently picks up the same proxy/PAC settings the OS and browsers use, falls back to direct connections only when no proxy exists, and reports the mode via telemetry. Outbound calls now honor their OS proxy without any manual settings.

## Future Work

This is a short-term solution to unblock customers. Long-term we should investigate a more robust solution for discovering proxies

## Testing

Tested by simulating customer environment on MacOS and Windows and using Charles Proxy:

Logs:

```
[Trace - 11:43:05 PM] Received notification 'window/logMessage'.
Params: {
    "type": 4,
    "message": "Fall back to OS auto‑detect (HTTP or HTTPS only)"
}


Fall back to OS auto‑detect (HTTP or HTTPS only)
[Trace - 11:43:05 PM] Received notification 'window/logMessage'.
Params: {
    "type": 4,
    "message": "os-proxy-config output: {\"proxyUrl\":\"http://127.0.0.1:8888\",\"noProxy\":[]}\n"
}


os-proxy-config output: {"proxyUrl":"http://127.0.0.1:8888","noProxy":[]}

[Trace - 11:43:05 PM] Received notification 'telemetry/event'.
Params: {
    "name": "runtime_httpProxyConfiguration",
    "result": "Succeeded",
    "data": {
        "proxyMode": "AutoDetect",
        "certificatesNumber": 328,
        "proxyUrl": "http://127.0.0.1:8888"
    }
}
```

Proxy:

Screenshot 2025-06-09 at 12 06 21 AM

## License

### DIFF
--- a/runtimes/runtimes/util/standalone/experimentalProxyUtil.ts
+++ b/runtimes/runtimes/util/standalone/experimentalProxyUtil.ts
@@ -13,6 +13,8 @@ import { NodeHttpHandler } from '@smithy/node-http-handler'
 import { readMacosCertificates, readLinuxCertificates, readWindowsCertificates } from './certificatesReaders'
 import { Telemetry } from '../../../server-interface'
 import { OperationalTelemetryProvider, TELEMETRY_SCOPES } from '../../operational-telemetry/operational-telemetry'
+import { execSync } from 'node:child_process'
+import { pathToFileURL } from 'node:url'
 
 export class ProxyConfigManager {
     /**
@@ -55,7 +57,7 @@ export class ProxyConfigManager {
         return this.cachedV3Config
     }
 
-    getProxyUrl(): string | undefined {
+    getProxyUrlFromEnv(): string | undefined {
         for (const envVar of this.PROXY_ENV_VARS) {
             const proxyUrl = process.env[envVar]
             if (proxyUrl) {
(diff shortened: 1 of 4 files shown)


runtimes/runtimes/util/standalone/experimentalProxyUtil.ts
```diff
@@ -69,6 +71,39 @@ export class ProxyConfigManager {
...
+                        const r   = await mod.getSystemProxy();
+                        console.log(JSON.stringify(r ?? {}));
+                    } catch (e) {
+                        console.error(e?.message ?? e);
+                        console.log("{}");
+                    }
+                })();
+            `
+
+            const raw = execSync(process.execPath, {
```
viktorsaws, line 92: Is this indirection needed? Would just calling `const proxy = await getSystemProxy();` work as expected?

PR #564, opened 2025-06-11
feat(types): add protocol for open file dialog for import image as context

yzhangok:
## Problem
To enable the import image in context feature, new protocol is added
Flow:
onOpenFileDialog(Chat client outbound) -> OpenFileDialog (Language-severs-runtime) -> language server -> lsp.window.showOpenDialog -> extension's API

After the user select an image

extension API -> language server -> OpenFileDialogResult (Language-severs-runtime) -> updatePrompt (Chat client inbound)

## Solution
`onOpenFileDialog`(Chat client outbound) -> `OpenFileDialog` (Language-severs-runtime) -> language server -> `lsp.window.showOpenDialog` -> extension's API

After the user select an image

extension API -> language server -> `OpenFileDialogResult` (Language-severs-runtime) -> updatePrompt (Chat client inbound)

## License

### DIFF
--- a/runtimes/protocol/chat.ts
+++ b/runtimes/protocol/chat.ts
@@ -73,6 +73,9 @@ import {
     McpServerClickParams,
     McpServerClickResult,
     MCP_SERVER_CLICK_REQUEST_METHOD,
+    OPEN_FILE_DIALOG_METHOD,
+    OpenFileDialogParams,
+    OpenFileDialogResult,
 } from './lsp'
 
 export const chatRequestType = new AutoParameterStructuresProtocolRequestType<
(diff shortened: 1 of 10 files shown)


runtimes/protocol/chat.ts
```diff
@@ -82,6 +85,13 @@ export const chatRequestType = new AutoParameterStructuresProtocolRequestType<
     void,
     void
 >(CHAT_REQUEST_METHOD)
+export const openFileDialogRequestType = new AutoParameterStructuresProtocolRequestType<
+    OpenFileDialogParams,
+    OpenFileDialogResult | string,
+    OpenFileDialogResult | string,
```
leigaol, line 91: why | string here?

yzhangok: For error message

zuoyaofu:
Approved for testing now. It is relatively safe since this PR is only adding new fields. @leigaol If you have comments, we can raise a PR to fix that.

PR #565, opened 2025-06-11
fix: bundle os-proxy-config and resolve it safely at runtime

tsmithsz:
## License


runtimes/runtimes/util/standalone/experimentalProxyUtil.ts
```diff
@@ -2,7 +2,7 @@
  * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
-
```
viktorsaws, line 5: I think `import { getSystemProxy } from 'os-proxy-config';` is intended usage of this package.

PR #576, opened 2025-06-20
feat(types): make the insertPosition optional in OpenFileDialogParams and OpenFileDialogResult

yzhangok:
## Problem
As the MynahUI code is refactored, change the protocol accordingly
 
## Solution

* Add additional field `content` for image context added by drag & drop to store the image content
* Make the `insertPosition` optional field

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -420,6 +420,7 @@ export interface ContextCommand extends QuickActionCommand {
     route?: string[]
     label?: 'file' | 'folder' | 'code' | 'image'
     children?: ContextCommandGroup[]
+    content?: Uint8Array
 }
 
 export interface ContextCommandParams {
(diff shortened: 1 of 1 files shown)


types/chat.ts
```diff
@@ -440,15 +441,15 @@ export interface CreatePromptParams {
 export interface OpenFileDialogParams {
     tabId: string
     fileType: 'image' | ''
-    insertPosition: number
+    insertPosition?: number
```
zuoyaofu, line 444: Just curious what is the reason to make it optional/required?

zuoyaofu (approved):

types/chat.ts
yzhangok: Refactor the code in MynahUI side

PR #577, opened 2025-06-22
fix: #512 change broke didSaveTextDocument handlers being called from the client

kddejong:
## Problem
- didSaveTextDocument is not working after the change https://github.com/aws/language-server-runtimes/pull/512.  While the testing looks good the `documents.listen` call hijacks the connection handlers.  https://github.com/microsoft/vscode-languageserver-node/blob/eafd0e224fc917f82f36ed19b6cb86bac77f9a98/server/src/common/textDocuments.ts#L173-L184

## Solution
Remove the observe and need for rxjs on TextDocuments.  This will result in relying fully on the Connection document calls.

## License


runtimes/runtimes/base-runtime.ts
```diff
@@ -317,7 +314,6 @@ export const baseRuntime = (connections: { reader: MessageReader; writer: Messag
         disposables.forEach(d => d())
     })
 
-    // Initialize the documents listener and start the LSP connection
-    documents.listen(documentsObserver.callbacks)
```
satyakigh, line 321: Same as above

kddejong: Yes `documents.listen` overrides the handler calls on the connection which results in `https://github.com/aws/language-server-runtimes/blob/35c19875d2b3429acbfbc8c5c348866499a3809e/runtimes/runtimes/standalone.ts#L377C17-L377C38` being void. The above line goes away because its no longer used.

There are alternative options here but based on the latest PR this fixes the issues and keeps to the standard that we seem to be moving to.

kmile: One of the reasons the `DocumentsObserver` was introduced was that while we want to be able to send all notifications to all `LspServer`s, we still want a single handler that makes sure document updates are processed in the right order, for the right `Uri`s, etc.

The code in `Workspace` at https://github.com/aws/language-server-runtimes/blob/eedc2603013c43479d76bbf523b06bafe333cf02/runtimes/runtimes/standalone.ts#L218-L219 relies on the `documents` repository to contain the up-to-date contents of all open documents.

`LpsServer`s themselves to not keep a duplicate copy of all documents, they just get a copy of the notifications to take any actions if necessary, but not to keep track of document state.

I think removing the `Observable` in favor of the `LspRouter` is great, it simplifies the dependencies and makes it consistent. You'd still need to attach the `TextDocuments` repository to the Lsp Connection though, and that needs to avoid conflict with the `LspRouter`. Maybe I missed that, but I can't find where that's happening in this PR (or what other mechanism replaces it).

kddejong: Yea, I'm not sure if it is getting replaced and that is a good catch.  The other option would be to undo the `onDidSaveTextDocument` change of #512 and revert it back to its original functionality which keeps the Observable pattern. 

I'm going to look a little bit more if there are other options to maintain both sides without reverting that change.  Right now we have pinned to `0.2.82` since we wanted the onDidSaveTextDocument notifications.

PR #578, opened 2025-06-23
fix: improve tool input validation logic to return specific error message

jguoamz:
## Problem
- When model passes incorrect type for the toolUse input, the error msg is not clear enough to allow auto correction
```
"toolResults": [
              {
                "toolUseId": "tooluse_f3iftfAAQ8uhRCHp1oJQCg",
                "status": "error",
                "content": [
                  {
                    "json": {
                      "error": "Input for tool fsReplace is invalid"
                    }
                  }
                ]
              }
            ]
```

## Solution
- improve tool input validation logic to return specific error message

## License

### DIFF
--- a/runtimes/runtimes/agent.ts
+++ b/runtimes/runtimes/agent.ts
@@ -1,4 +1,4 @@
-import Ajv from 'ajv'
+import Ajv, { ErrorObject, ValidateFunction } from 'ajv'
 import {
     Agent,
     BedrockTools,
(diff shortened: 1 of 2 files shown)


runtimes/runtimes/agent.ts
```diff
@@ -48,7 +55,12 @@ export const newAgent = (): Agent => {
             }
 
             if (!tool.validate(input, token)) {
-                throw new Error(`Input for tool ${toolName} is invalid`)
+                const errorDetails =
```
ctlai95, line 58: I see you're setting the validationError outside of this scope at the tool level to reuse here, but it might be cleaner if we can just return the error from `tool.validate()`?

jguoamz: tool.validate() will not return any error, the error is from the validator

ctlai95: yes but we control what tool.validate returns

jguoamz: fixed

ctlai95 (approved):

ctlai95:
We will need to update this error mapper to include this new error message

https://github.com/aws/language-servers/blob/main/server/aws-lsp-codewhisperer/src/language-server/agenticChat/errors.ts#L243

PR #585, opened 2025-06-26
fix: remove duplicate uri of workspace paths sent by VS

XiaoxuanLu:
## Problem
Visual Studio was sending duplicate workspace folder URIs in workspace change events, which caused duplicate rules to appear when listing rules.

## Solution
Added deduplication filter to check existing workspace folder URIs before adding new ones, preventing duplicates from being stored in the workspaceFolders array.

## License

### DIFF
--- a/runtimes/runtimes/lsp/router/lspRouter.ts
+++ b/runtimes/runtimes/lsp/router/lspRouter.ts
@@ -144,7 +144,9 @@ ${JSON.stringify({ ...result.capabilities, ...result.awsServerCapabilities })}`
         this.workspaceFolders = this.workspaceFolders.filter(
             folder => !event.removed.some(removed => removed.uri === folder.uri)
         )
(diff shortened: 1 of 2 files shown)


runtimes/runtimes/lsp/router/lspRouter.ts
```diff
@@ -144,7 +144,9 @@ ${JSON.stringify({ ...result.capabilities, ...result.awsServerCapabilities })}`
         this.workspaceFolders = this.workspaceFolders.filter(
             folder => !event.removed.some(removed => removed.uri === folder.uri)
         )
-        this.workspaceFolders.push(...event.added)
+        this.workspaceFolders.push(
+            ...event.added.filter(added => !this.workspaceFolders.some(existing => existing.uri === added.uri))
```
awschristou, line 148: Can we add a test for this

awschristou (approved):

PR #589, opened 2025-06-27
chore: add timeout to ensure error flushes to stdout/err on auth failure

bryceitoc9:
## Problem
No log output for initializeAuth failure

## Solution
Add a timeout before process kill to ensure the buffer flushes

Tested locally by directly modifying the prod artifact. Note that the webpacked version appears to only output to `stdout` (instead of `stderr`, which should be expected with a `console.error`

## License

### DIFF
--- a/runtimes/runtimes/standalone.ts
+++ b/runtimes/runtimes/standalone.ts
@@ -184,7 +184,11 @@ export const standalone = (props: RuntimeProps) => {
                     },
                     error => {
                         console.error(error)
-                        process.exit(10)
+                        // arbitrary 5 second timeout to ensure console.error flushes before process exit
+                        // note: webpacked version may output exclusively to stdout, not stderr.
+                        setTimeout(() => {
+                            process.exit(10)
+                        }, 5000)
                     }
                 )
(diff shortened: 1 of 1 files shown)


runtimes/runtimes/standalone.ts
```diff
@@ -184,7 +184,11 @@ export const standalone = (props: RuntimeProps) => {
                     },
                     error => {
                         console.error(error)
-                        process.exit(10)
+                        // arbitrary 5 second timeout to ensure console.error flushes before process exit
```
awschristou, line 187: Wouldn't you want to force-flush the console output and exit immediately? Waiting 5 more seconds seems like a side effect, one where a client could start sending what it intended to during the 2 second window of time the system was listening for data, only to have the process terminate.
```
process.stdout.write()
process.stderr.write()
```

awschristou (approved):

runtimes/runtimes/standalone.ts
rli: `process.stderr.flush()`?

PR #595, opened 2025-07-01
fix: refactor auto-discovery for OS proxy settings

tsmithsz:
# Problem

Related PR:
- https://github.com/aws/language-server-runtimes/pull/553

We originally bolted on an exec-time shim that `require.resolve’d` [os-proxy-config](https://www.npmjs.com/package/os-proxy-config) and ran it in a child-process so we could keep `ProxyConfigManager.getSystemProxySync()` fully synchronous. This avoided a ripple-effect async refactor through every SDK-initialisation call-chain and let us ship quickly, but it also meant the code was never picked up by Webpack, so the proxy lookup silently no-oped inside the bundled LSP.

## Solution

- Vendored the tiny os-proxy-config implementation directly  (`getMacProxySettings.ts`, `getWindowsProxySettings.ts`) and removed the child-process shim.
   - Reference Code: [os-proxy-config](https://github.com/httptoolkit/os-proxy-config), [windows-system-proxy](https://github.com/httptoolkit/windows-system-proxy/tree/main), [mac-system-proxy](https://github.com/httptoolkit/mac-system-proxy/tree/main)
- Re-wrote the Windows & macOS helpers to expose sync variants (registry reads with `registry-js` sync API, `scutil --proxy` via `spawnSync`) so `getSystemProxy()` remains blocking.
- Behaviour is unchanged for callers, but proxy auto-detect now runs in all packed LSPs because the code is part of the bundle.

## TODO
- Add support for SOCKS Proxy

## Testing

Simulated customer proxy environments on Windows and Mac and tested plugin build:

- #### Mac

Proxy is being auto-detected and requests are going through proxy:

Screenshot 2025-07-01 at 12 44 15 AM

- #### Windows

Proxy is being auto-detected and requests are going through proxy:

```
Using bare hostname format, defaulting to HTTP
[Trace - 6:30:06 PM] Received notification 'telemetry/event'.
Params: {
    "name": "runtime_httpProxyConfiguration",
    "result": "Succeeded",
    "data": {
        "proxyMode": "AutoDetect",
        "certificatesNumber": 208,
        "proxyUrl": "http://localhost:8888/"
    }
}
```

Screenshot 2025-07-01 at 6 32 53 PM

## License

### DIFF
--- a/runtimes/runtimes/util/standalone/getProxySettings/parseScutil.ts
+++ b/runtimes/runtimes/util/standalone/getProxySettings/parseScutil.ts
@@ -1,0 +1,60 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Based on mac-system-proxy 1.0.2 (Apache-2.0).
+ * https://github.com/httptoolkit/mac-system-proxy/blob/main/src/parse-scutil.ts
+ */
+const TYPE_KEY = '__scutil__type__'
+
+// Quick hacky parser, which translates output into valid JSON:
+export function parseScutilOutput(output: string): {} {
+    try {
+        // Unclear how this happens, but it seems that it can in some cases:
+        if (output === '') return {}
+
+        const unquotedJsonString = output
+            // Reduce type markers to just an inline __scutil__type__ marker on array objects:
+            .replace(/<dictionary> /g, '')
+            .replace(/<array> {/g, `{\n${TYPE_KEY} : array`)
+            .trim()
+
+        // Turn unquoted key/value string into a string of quote key values (but with no commas).
+        // We effectively parse by splitting on the first " : " in each line, if present.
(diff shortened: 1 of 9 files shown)


samgst-amazon (approved):

runtimes/runtimes/util/standalone/getProxySettings/parseScutil.ts
```diff
@@ -0,0 +1,57 @@
+/*!
+ * Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+const TYPE_KEY = '__scutil__type__'
+
+// Quick hacky parser, which translates output into valid JSON:
+export function parseScutilOutput(output: string): {} {
```
leigaol, line 8: can you find a library that does this for you? This JSON string manipulation is fragile

tsmithsz: Discussed offline

PR #602, opened 2025-07-07
fix(runtimes): timeout for exiting to emit logs

samgst-amazon:
## Problem
LSP was exiting on Uncaught Exception before the error message could be logged, making it difficult to debug root cause
## Solution
Add a timeout for the exit process to let the server emit telemetry and log the error before exiting

## License


rli (approved):

bryceitoc9 (approved):
Discussed offline, looks like the `uncaughtExceptionMonitor` is handling output.

rli (approved):

leigaol (approved):

PR #606, opened 2025-07-11
feat(types): add keybinding type for shortcut

dungdong-aws:
## Problem

Handle keyboard shortcut for stop/run/reject shell-command

## Solution

Checking for keybinding for the shortcuts then add to tooltip before rendering `Reject/Run` button for shell-commands

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
     models: Model[]
     selectedModelId?: string
 }
+
+export interface KeyBinding {
+    key: string
+    command: string
+    when?: string
+}
+
+export interface KeybindingResult {
+    key: string | null
+    isExplicitlyDisabled: boolean
+    isInKeybindingsFile: boolean
+}
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -693,11 +693,4 @@ export interface ListAvailableModelsResult {
 export interface KeyBinding {
     key: string
     command: string
-    when?: string
-}
-
-export interface KeybindingResult {
-    key: string | null
-    isExplicitlyDisabled: boolean
-    isInKeybindingsFile: boolean
 }


types/chat.ts
```diff
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
...
+    key: string
+    command: string
+    when?: string
+}
+
+export interface KeybindingResult {
+    key: string | null
+    isExplicitlyDisabled: boolean
+    isInKeybindingsFile: boolean
+}
```
yueny2020, line 703: could you please explain how we will use these? thanks.

types/chat.ts
```diff
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
...
 }
+
+export interface KeyBinding {
+    key: string
+    command: string
+    when?: string
+}
+
+export interface KeybindingResult {
+    key: string | null
```
ashishrp-aws, line 700: use optional value of string instead of null

types/chat.ts
```diff
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
     models: Model[]
     selectedModelId?: string
 }
+
+export interface KeyBinding {
+    key: string
+    command: string
+    when?: string
```
ashishrp-aws, line 696: why is this called `when`? can you name it something general?

types/chat.ts
```diff
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
...
+}
+
+export interface KeybindingResult {
+    key: string | null
+    isExplicitlyDisabled: boolean
```
ashishrp-aws, line 701: don't use explicitily

types/chat.ts
```diff
@@ -689,3 +689,15 @@ export interface ListAvailableModelsResult {
...
+    isInKeybindingsFile: boolean
```
ashishrp-aws, line 702: Don't make the option naming specific to your usecase

ashishrp-aws (approved):

types/chat.ts
yueny2020: `keybinding.json` is a file you will created for this feature?

dungdong-aws: it's from VSC, if you change a keybind shortcut for something, it will be defined in that file. It's generated by VSC

yueny2020: can we have some listener for the keybinding change?

dungdong-aws: Im not sure, pretty sure we cant? because it's from VSC side, they suggest to check for that file instead

yueny2020: > check for that file
could be file change listener?

dungdong-aws: Because I think it will be easier to just grab it on the go when we render Reject/Run buttons.

yueny2020: if you don't have listener, then you want to read the json file before every time the button rendering?

dungdong-aws: >  file change listener?

user could change it offline without open IDE, if it the case, how can we listen for it?

dungdong-aws: > if you don't have listener, then you want to read the json file before every time the button rendering?

Yes, that file is only written if user change default key of a command. So it wont be a long json

PR #650, opened 2025-08-04
feat: encrypt updateProfile and listProfiles endpoints

liramon1:
## Problem
The IAM authentication changes assume that the language server and language client exchange IAM credentials in plaintext through the listProfiles and updateProfile endpoints. Sensitive data, like IAM credentials, should instead be encrypted in-transit to follow best security practices.

## Solution
- Encrypt profiles returned by listProfiles
- Decrypt profiles sent to updateProfile

**This is a breaking change since language clients must decrypt listProfiles responses**

## License

### DIFF
--- a/runtimes/protocol/identity-management.ts
+++ b/runtimes/protocol/identity-management.ts
@@ -141,7 +141,7 @@ export interface ListProfilesResult {
 // Potential error codes: E_UNKNOWN | E_TIMEOUT | E_RUNTIME_NOT_SUPPORTED | E_CANNOT_READ_SHARED_CONFIG
 export const listProfilesRequestType = new ProtocolRequestType<
     ListProfilesParams,
-    ListProfilesResult,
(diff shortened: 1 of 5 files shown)


runtimes/protocol/identity-management.ts
```diff
@@ -141,7 +141,7 @@ export interface ListProfilesResult {
 // Potential error codes: E_UNKNOWN | E_TIMEOUT | E_RUNTIME_NOT_SUPPORTED | E_CANNOT_READ_SHARED_CONFIG
 export const listProfilesRequestType = new ProtocolRequestType<
     ListProfilesParams,
-    ListProfilesResult,
+    ListProfilesResult | string,
```
awschristou, line 144: Can we avoid having two different types of responses? Clients written in different languages will receive a json blob, and cannot reliably differentiate a structured type from a string.

PR #655, opened 2025-08-06
feat: editable variable for modify functionality 

qjhuangAWS:
## Problem
The current chat interface lacks support for editing commands/messages after they have been sent. This limits user experience when users need to modify their chat prompts or correct mistakes without starting a new conversation thread.

## Solution
Added support for message editing by extending the chat type definitions: 
add editable property to ChatMessage and editedText to ButtonClickParams

## License

### DIFF
--- a/types/chat.ts
+++ b/types/chat.ts
@@ -404,6 +404,9 @@ export interface ButtonClickParams {
     messageId: string
     buttonId: string
     metadata?: Record<string, string>
+    /**
+     * Enables passing edited text content when users interact with edit-related buttons
+     **/
     editedText?: string
 }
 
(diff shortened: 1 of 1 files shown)


types/chat.ts
```diff
@@ -403,6 +404,10 @@ export interface ButtonClickParams {
     messageId: string
     buttonId: string
     metadata?: Record<string, string>
+    /**
+     * Enables passing edited text content when users interact with edit-related buttons
```
ctlai95, line 408: Could this be solved using `metadata` ?

qjhuangAWS: yep! metadata? was also newly created, did't realize it existed

ctlai95 (approved):

PR #425, opened 2025-04-11
chore: move aws-sdk to devDependencies

rahmaniaam:
## Problem
https://github.com/aws/language-server-runtimes/issues/417

`aws-sdk` adds significant size to the package (103MB). Since it's only used for types declaration, we can move it to devDependencies.

Thanks @TrevorBurnham for reporting and immediately giving out the solution!

## Solution
Move `aws-sdk` to devDependencies

## License

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #417, opened 2025-04-08
@aws/language-server-runtimes depends on deprecated aws-sdk v2


TrevorBurnham:
If you install the latest `@aws/language-server-runtimes` with visible postinstall script output (e.g. using npm's `--foreground-scripts` flag), you'll see this:

```
╔═════════════════════════════════════════════════╗
║ The AWS SDK for JavaScript (v2) will reach      ║
║ -> maintenance mode on September 8, 2024.       ║
║ -> end-of-support on September 8, 2025.         ║
║                                                 ║
║ To continue receiving updates to AWS services,  ║
║ bug fixes, and security updates please upgrade  ║
║ to AWS SDK for JavaScript (v3).                 ║
║                                                 ║
║ More info: https://a.co/cUPnyil                 ║
╚═════════════════════════════════════════════════╝
```

Sure enough, the package [directly depends on aws-sdk v2](https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/package.json#L49).

Aside from the deprecation notice, using aws-sdk v2 adds a huge amount of code to downstream projects. The size of my `node_modules/aws-sdk` is 103 MB!

Interestingly, looking around the code it looks like this package is only used for types, so the simplest fix would be to move `aws-sdk` into `devDependencies`.

If I understand [this code](https://github.com/aws/language-server-runtimes/blob/674c02696c150838b4bc93543fb0009c5982e7ad/runtimes/runtimes/standalone.ts#L395-L429) correctly, the purpose of those types is backward compatibility with aws-sdk v2, so an alternative option would be to release a new version that drops that backward compatibility.

justinmk3:
> looking around the code it looks like this package is only used for types, so the simplest fix would be to move `aws-sdk` into `devDependencies`.

Great idea!

PR #668, opened 2025-08-23
perf: replace aws-sdk with @types/aws-sdk2-types

TrevorBurnham:
This is a second attempt at fixing #417. The first attempt was #425, which moved `aws-sdk` from `dependencies` to `devDependencies`. That change was reverted by #535.

The `aws-sdk` dependency is only being used for TypeScript types. Including it in `dependencies` rather than `devDependencies` ensures that those types are available to package consumers, at the expense of bringing in a 100MB+ dependency.

The solution here is to replace the `aws-sdk` dependency with `@types/aws-sdk2-types`, which is a relatively lightweight 2.3MB. These types should be identical, as they were pulled directly from `aws-sdk`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64111

## Further potential optimizations

I considered inlining the types with [dts-bundle-generator](https://github.com/timocov/dts-bundle-generator), which would allow `@types/aws-sdk2-types` to be moved to `devDependencies`, but this turned out to be impractical with the current export structure, since the types would have to be inlined separately in each `.d.ts` file that uses them. I would recommend that solution if the package moves to having a single defined entrypoint in the future (perhaps in the next major version).

## License


chungjac:
Fixed in https://github.com/aws/language-server-runtimes/pull/685

